PR #373, opened 2023-06-23
Update load tests to read config

fkrause98:
## Changes:
- Fix load tests, they were broken because of out-of-date config.
## How to test:
- Follow the readme from the load test app


tvillegas98 (approved):
LGTM!

PR #376, opened 2023-06-23
Use Flat Vector for Grid.

fkrause98:
## Changes:
- Start using a Flat Vector instead of a Vector of Vectors for the grid internal structure, as to avoid cache misses and achieve a more cpu-friendly data structure.
- TL;DR It should be more performant this way.
## Mini-report:

- There is an expected peak of CPU usage when multiple connections start, then it stabilizes, and 
  then the general pattern seems to be an improvement on CPU% reported by htop on the flat vec version.
- I've managed to run 40 games with 30 players.
- I've noticed that when I start 45+ games with 30 players each the server starts to run out of file descriptors.
(Maybe this is due to TCP sockets being used by Websocket?)
- Here are some relevant htop captures from the load tests I did on Friday 23th June.
### One Game with 30 players, with a flat Vec:
FlatVec - One Game 30 players

### One Game with 30 players, with a nested Vec:
Main -One Game - 30 players

### 10 Games with 30 players, flat Vec:
FlatVec - 10 Games - 30 players

### 10 Games with 30 players, with a nested Vec:
Main - 10 Games - 30 players


server/native/gamestate/src/board.rs
```diff
@@ -1,9 +1,13 @@
...
-
+pub type Grid = Vec<Tile>;
 #[derive(Debug)]
 pub struct GridResource {
-    pub resource: Mutex<Vec<Vec<Tile>>>,
+    pub resource: Mutex<Grid>,
+}
+#[derive(Debug)]
+pub struct FlatGridResource {
+    pub resource: Mutex<Vec<Tile>>,
```
AminArria, line 10: Shouldn't this be 

```rust
pub resource: Mutex<Grid>
```

server/native/gamestate/src/board.rs
```diff
@@ -33,17 +37,36 @@ impl Board {
...
-            row.get(col_idx).cloned()
-        } else {
-            None
-        }
+        let indx = (row_idx * self.width) + col_idx;
+        self.grid
+            .resource
+            .lock()
+            // I don't really like this, but it hasn't showed
+            // up in load tests, so I guess this is ok.
```
AminArria, line 46: What don't you like?

server/native/gamestate/src/board.rs
```diff
@@ -1,9 +1,13 @@
 use rustler::{NifStruct, NifTaggedEnum, ResourceArc};
 use std::sync::Mutex;
-
+pub type Grid = Vec<Tile>;
 #[derive(Debug)]
 pub struct GridResource {
```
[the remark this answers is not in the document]
fkrause98: We do not, forgot to erase it. Removed.

server/native/gamestate/src/board.rs
fkrause98: I don't like having unwraps/expects, but that line of code hopefully always checks out as the BEAM schedules function calls properly.

AminArria: What alternative is there to avoid the `expect`?

fkrause98: The alternative is to return a result and simply 'bubble up' the error but this poses some questions:
- Should this be somehow recoverable? A lock error means something went wrong with the scheduling of the BEAM, what should we do if this error is reported?
- Shouldn't we abort everything if this happnes?

jrchatruc (approved):

tvillegas98 (approved):

tvillegas98:
We will merge it once @AminArria finished his discussion with @fkrause98 😄

AminArria (approved):

PR #387, opened 2023-06-26
Coordiantes refactor

### LINKED ISSUE #386: [REFACTOR] Unify JoystickValues

manucamejo:
Closes #386

### DIFF
--- a/server/native/gamestate/src/game.rs
+++ b/server/native/gamestate/src/game.rs
@@ -176,7 +176,7 @@ impl GameState {
         direction: &RelativePosition,
     ) -> Result<(), String> {
         // TODO: 120 should be a config. It's the realtion between front range in skills and
-        // the distance in the back. 
+        // the distance in the back.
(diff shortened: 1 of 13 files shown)


server/native/gamestate/src/game.rs
```diff
@@ -363,25 +367,25 @@ impl GameState {
     }
 
     pub fn position_to_direction(position: &RelativePosition) -> Direction {
```
fkrause98, line 369: I would move this function to where RelativePosition was defined, we have too many functions already here.

fkrause98: Even more so, we could make a From<Direction> for RelativePosition trait to simply use .into() when we want to do this type transformation.

manucamejo: we are going to refactor this code with the new distance calculation

tvillegas98 (approved):
Seems to be working as it should 😄

jrchatruc (approved):

PR #388, opened 2023-06-26
Bot movement

### LINKED ISSUE #355: [FEATURE] Add agressive bot
Add a simple AI to bots that make them move and try to attack adjacent players

AminArria:
Part of #355 

- [x] Bot moves randomly
- [x] Button to enable/disable bot actions

In next PR
- [ ] Bot moves random or towards player
- [ ] Bot attacks nearby players
- [ ] Random selection of character for bots

### DIFF
--- a/client/Assets/Prefabs/Camera/CameraUI.prefab
+++ b/client/Assets/Prefabs/Camera/CameraUI.prefab
@@ -2835,6 +2835,7 @@ RectTransform:
   - {fileID: 456601721}
   - {fileID: 8589184727428656987}
   - {fileID: 8589184728810921813}
+  - {fileID: 4549140611734199729}
   m_Father: {fileID: 4776084761794064}
   m_RootOrder: 1
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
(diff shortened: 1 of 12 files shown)


BertovDev:
Nice ! I think it would be good to add a toggle button for the bot movement. We can do that in other PR

ncontinanza (changes requested):
Nice job! Just one detail I noticed: while the bots are moving, it is impossible to damage them by attacking, and they do not perform the animation of receiving an attack.

AminArria:
>  toggle button for the bot movement.

@BertovDev added it to my todo list, for now I also included a "no movement" in the random so it can happen they stay still

> impossible to damage them by attacking

@ncontinanza seems this is related to how fast the bot gets and sends updates vs the client, I think the bot is just too in the future or the client is too in the past and we have no way to compensate for this. For now I slowed the bot a bit (further slowed would mess with the movement for some reason) and it seems more damageable, let me know how you feel it

ncontinanza (approved):

BertovDev (commented):
Awesome! The only thing is that the bots are pretty difficult to hit. ( It simulate real game behaviour really well which is awesome ) but it might be a disadvantages for quick testings.

I think it would be better to merge it when the no movement feature is implemented.
What do you think?

tvillegas98:
> Awesome! The only thing is that the bots are pretty difficult to hit. ( It simulate real game behaviour really well which is awesome ) but it might be a disadvantages for quick testings.
> 
> I think it would be better to merge it when the no movement feature is implemented. What do you think?

Maybe what we can add are buttons to toggle on the movement or the attack

tvillegas98:
This is working very well, I test it with 9 bots and each one of them moved around the map 😛  and also, I was able to hit each one of them

tvillegas98 (approved):

BertovDev (approved):
Toggle Bot button works perfect, so nice!

PR #392, opened 2023-06-27
Process priority benchmark

fkrause98:
(Wri

### DIFF
--- a/server/lib/dark_worlds_server/engine/runner.ex
+++ b/server/lib/dark_worlds_server/engine/runner.ex
@@ -74,6 +74,7 @@ defmodule DarkWorldsServer.Engine.Runner do
       Application.fetch_env!(:dark_worlds_server, __MODULE__)
       |> Keyword.fetch!(:process_priority)
 
+    Logger.info("Starting game with priority: #{priority}")
     Process.flag(:priority, priority)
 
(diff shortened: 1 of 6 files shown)


server/lib/dark_worlds_server/engine/runner.ex
```diff
@@ -457,7 +458,7 @@ defmodule DarkWorldsServer.Engine.Runner do
 
   defp all_characters_set?(state) do
     cond do
-      state[:game_status] == :playing ->
+      Map.get(state, :game_status) == :playing ->
```
AminArria, line 461: Any reason or just preferences?

docs/src/benchmarking.md
```diff
@@ -0,0 +1,45 @@
+## Benchmarking guide
+For benchmarking, you should follow this steps:
+1. Deploy your branch, obviously.
```
AminArria, line 3: You would be surprise how many people forget this 😂

AminArria (approved):
LGTM

server/lib/dark_worlds_server/engine/runner.ex
tvillegas98: It was causing an error 🤷‍♂️ . We changed it for that and it worked 😆

jrchatruc (approved):

PR #395, opened 2023-06-27
Left joystick UX changes

AngieDutra:
This PR adds a new feature for the left Joystick
The user can press in certain section and define the joysticks position before using it. Then, it will go back to its original position. It also has a logic to prevent the joystick from going out of bounds and stay in screen.


saugon (approved):
Niiice job! It works like a charm :) 🚀

BertovDev (approved):
Stunnish Work!

PR #399, opened 2023-06-28
Skills refactor + Animations

saugon:


client/Assets/Scripts/PlayerMovement.cs
```diff
@@ -37,25 +47,59 @@ void Update()
             UpdatePlayerActions();
             UpdateProyectileActions();
         }
+        GameObject player = Utils.GetPlayer(SocketConnectionManager.Instance.playerId);
```
BertovDev, line 50: This `player` is not been used , right ? and its been assing in every frame. Should be deleted.

client/Assets/Scripts/PlayerMovement.cs
```diff
@@ -262,15 +332,22 @@ private void movePlayer(GameObject player, Player playerUpdate)
         mAnimator.SetBool("Walking", walking);
 
         Health healthComponent = player.GetComponent<Health>();
-        // Display damage done on you on your client
-        GetComponent<PlayerFeedbacks>().DisplayDamageRecieved(player, healthComponent, playerUpdate.Health, playerUpdate.Id);
 
-        // Display damage done on others players (not you)
-        GetComponent<PlayerFeedbacks>().ChangePlayerTextureOnDamage(player, healthComponent.CurrentHealth, playerUpdate.Health);
+        // FIXME: Temporary solution until all models can handle the feedback
+        if (playerUpdate.CharacterName == "H4ck"){
```
BertovDev, line 337: This if is not working. It throw me an error  because `playerUpdate.CharacterName` return "H4ck" also using Muflus , it can be fixed by replacing it with :

`player.GetComponent<Character>().CharacterModel.name == "H4ck"`

client/Assets/Scripts/PlayerMovement.cs
```diff
@@ -262,15 +332,22 @@ private void movePlayer(GameObject player, Player playerUpdate)
...
+            // Display damage done on you on your client
+            GetComponent<PlayerFeedbacks>().DisplayDamageRecieved(player, healthComponent, playerUpdate.Health, playerUpdate.Id);
```
BertovDev, line 339: This is not necessary to be inside the if condition because it doesnt matter the type of character. It just need to have the `DamageFeedback` gameObject ( which is missing in Muflus)

client/Assets/Scripts/PlayerMovement.cs
```diff
@@ -262,15 +332,22 @@ private void movePlayer(GameObject player, Player playerUpdate)
...
+
+            // Display damage done on others players (not you)
+            GetComponent<PlayerFeedbacks>().ChangePlayerTextureOnDamage(player, healthComponent.CurrentHealth, playerUpdate.Health);
```
BertovDev, line 342: Same with this, dont need to be in the if statement , just need to have the `DeathFeedback `gameObject which is also missing.

BertovDev (changes requested):
This is a really good job, there are some corrections to do in Muflus. But everything else looks clean! I will list the following the fixes to do in Muflus ( we can do it in another PR ).

- Set the min health to 100 in the health component ( To avoid damage feedback bugs )
- Add DamageFeedback Gameobject
- Add DeathFeedback Gameobject
- Fix DisplayName position

client/Assets/Scripts/PlayerMovement.cs
BertovDev: My bad , this return H4ck when you use Spawn a Bot using the Muflus Model ( For testing purposes) , so we can use` playerUpdate.CharacterName` but its not going to work if we want to test a Bot with muflus character

client/Assets/Scripts/PlayerMovement.cs
saugon: Done! Thanks :)

client/Assets/Scripts/PlayerMovement.cs
saugon: Done! Thanks :)

ncontinanza (approved):

BertovDev (approved):
Awesome!

PR #403, opened 2023-06-29
Fix runner tickrate

tvillegas98:
We were not getting the tickrate from the game config, instead we were always using the default one.


jrchatruc (approved):

manucamejo (approved):
LGTM

PR #407, opened 2023-06-29
Add overlay shader to display Damage

BertovDev:
This PR adds a Shader to display the damage Feedback in the player. Using an Overlay shader we fix the problem of characters having differents materials and model hierarchy.

With this you only have to attach to the OverlayShader prefab the SkinnedMeshRenderer component of the model, in case it has multiple SkinnedMeshRenderer components attach the parent which contains all of them.


saugon (approved):
Nice work! 
Disclaimer: in the future we'll change the visuals but in the meantime it's super useful.

tvillegas98 (approved):

PR #415, opened 2023-07-03
Right panel UX

AngieDutra:
This PR:

- adds new assets according to the player selected
- prevents the joystick from be used if its disabled
- the joysticks load when the sprites are set
- if the user presses the joystick button without dragging it, it no longer activates

Notes:
"if the user presses the joystick button without dragging it, it no longer activates"
- however, it stays with an opacity as if it were pressed, i will fix this in another PR


tvillegas98 (approved):
LGTM!

BertovDev (approved):
Tested in pc ! LGTM

ncontinanza (approved):
LGTM! Tested it with an Android build, works like a charm.

PR #480, opened 2023-07-05
Migrate unity project to URP

BertovDev:
In this PR we implement the Universal Render Pipeline as default pipeline , also convert all the materials to URP lit and add VFX compatible with URP.

- [x] Android build test
- [x] iOS build test

### IMPORTANT

-  Before making a build to test this PR you have to delete the APK you already have in your phone ( if you have ) if is not deleted it will throw an error in the console.
-  In the first run you might experience some broken textures.Ones your unity detecs the new materials it will be shown correctly


tvillegas98:
Uploading Grabación de pantalla 2023-07-07 a la(s) 11.32.23.mov…


H4ck's damage feedback is a bit buggy 😢 . Also the damage feedback that's currently being show as a flashy red screen is not working

manucamejo (approved):

PR #482, opened 2023-07-06
Load skills from Skills.json

### LINKED ISSUE #431: [FEATURE] Setup skills from configuration file
## 🚀 Feature

We should read `Skills.json` and use it to setup the parameters of the skills in the backend and client

1. Parse `Skills.json` in client and send it to backend
2. Backend uses skills config to modify parameters of skills (damage, cooldown, etc)
3. Client uses skills config to modify parameters of skills

AminArria:
Closes #431 

Main changes
- Skills and their parameters are loaded from `Skills.json`
- Now we have a single `Skills` struct, I don't see any reason to limit the skills to be tied to a specific type (e.g. basic, skill2, etc)


client/Assets/Scripts/GameSettings.cs
```diff
@@ -45,20 +53,33 @@ public static IEnumerator ParseSettingsCoroutine(Action<ServerGameSettings> call
                     {
                         jsonCharacterSettingsText = charactersResult;
 
-                        RunnerConfig parsedRunner = parser.Parse<RunnerConfig>(
-                            jsonGameSettingsText.TrimStart('\uFEFF')
-                        );
-                        CharacterConfig characters = parser.Parse<CharacterConfig>(
-                            jsonCharacterSettingsText.TrimStart('\uFEFF')
-                        );
+                        GetRequest(
```
AminArria, line 56: Tried to avoid this weird nesting thing, but couldn't find a way (in the hour I took), so will leave it for a next refactor (a.k.a when we add more configs)

server/native/gamestate/src/skills.rs
```diff
@@ -1,78 +1,86 @@
...
-    )]
-    BarrelRoll,
-    #[strum(serialize = "Serpent Strike", serialize = "SerpentStrike")]
-    SerpentStrike,
-    #[strum(ascii_case_insensitive)]
-    MultiShot,
-    #[strum(serialize = "Yugens Mark", serialize = "Pizzicato")]
-    Placeholder,
+impl Default for Skill {
+    fn default() -> Self {
```
AminArria, line 45: The only reason we need this is because the functions doing the bot spawn are using `Character::default()`

server/native/gamestate/src/character.rs
```diff
@@ -1,9 +1,21 @@
...
-use crate::skills::{Basic as BasicSkill, Class, FirstActive, SecondActive};
+use rustler::NifTaggedEnum;
 use std::collections::HashMap;
+use std::ops::Div;
 use std::str::FromStr;
 use strum_macros::{Display, EnumString};
 pub type TicksLeft = u64;
+
+#[derive(NifTaggedEnum, Debug, Clone, EnumString, Display)]
+pub enum Class {
```
AminArria, line 10: Moved here from `skills.rs`

server/native/gamestate/src/character.rs
```diff
@@ -1,9 +1,21 @@
...
+    #[strum(serialize = "hun", serialize = "Hunter", ascii_case_insensitive)]
+    Hunter,
+    #[strum(serialize = "gua", serialize = "Guardian", ascii_case_insensitive)]
+    Guardian,
+    #[strum(serialize = "ass", serialize = "Assassin", ascii_case_insensitive)]
+    Assassin,
+}
```
AminArria, line 17: Moved here from `skills.rs`

saugon:
@AminArria could we add the attack range to the json?

This should affect for example Muflus -> Barrel Roll skill, that is hardcoded in the code.

AminArria:
> @AminArria could we add the attack range to the json?
> This should affect for example Muflus -> Barrel Roll skill, that is hardcoded in the code.

After discussing, will be done in a follow up PR to close this one @saugon

client/Assets/Scripts/GameSettings.cs
tvillegas98: I agree, this looks weird but we should find a way to fix this later

server/native/gamestate/src/lib.rs
```diff
@@ -20,22 +21,11 @@ fn new_game(
     board_width: usize,
     board_height: usize,
     build_walls: bool,
-    characters_config: Vec<HashMap<Binary, Binary>>,
+    raw_characters_config: Vec<HashMap<Binary, Binary>>,
```
tvillegas98, line 24: Thanks for this!

tvillegas98 (approved):
LGTM!

manucamejo (approved):

PR #497, opened 2023-07-10
Fix character selection bug

### LINKED ISSUE #489: [BUG] Character selection
Related:https://github.com/lambdaclass/curse_of_myrra/issues/447

If we select a character too quickly in the character selection, the server doesn’t receive the message, even if we select again the character

tvillegas98:
closes #489

The bug happens when we select a character too quickly that the game didn't even connected to the server, thus, our character selection breaks and the entire game too.
In this PR, I introduced a new function called `isConnectionOpen` in the `SocketConnectionManager`, maybe in the future we can use that function for loading screens.

### DIFF
--- a/client/Assets/Scripts/LobbyConnection.cs
+++ b/client/Assets/Scripts/LobbyConnection.cs
@@ -308,4 +308,9 @@ public class LobbyConnection : MonoBehaviour
             return "wss://" + server_ip + path;
         }
     }
+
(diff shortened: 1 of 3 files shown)


BertovDev (approved):
FIxes the bug,  Awesome!!.

saugon (approved):
Nice!

Issue #448, opened 2023-07-05
[BUG] Game state breaks after returning to lobbies


saugon:
We have seen different errors once the player returns to Lobbies screen from a match and joins a new one.
For example, character rotation stops working.

PR #508, opened 2023-07-10
Fix SocketConnectionManager Singleton

ncontinanza:
Fixes some of the issues related to #448

The SocketConnectionManager is intended to be a Singleton object, but until now it was possible to have more than one instance when finishing a game and starting a new one.

Also, a back button was added to the character selection scene.

## Steps to test:
1. Launch a new game in your editor. Check that you have only one instance of the NetworkManager in the character selection scene, that is also alive in the BackendPlayground scene.
2. Pause the game and go back to the lobbies scene. Verify that the NetworkManager was destroyed in this step.
3. Launch a new game, and you should see only one instance of the NetworkManager in the character selection scene.

When you are in the character selection scene, you can also test the back button, that should take you back to the lobbies scene.


client/ProjectSettings/ProjectSettings.asset
```diff
@@ -255,8 +255,8 @@ PlayerSettings:
   AndroidTargetDevices: 0
   AndroidSplashScreenScale: 0
   androidSplashScreen: {fileID: 0}
-  AndroidKeystoreName: /Users/bautistaberto/lambda/user.keystore
-  AndroidKeyaliasName: bauti
+  AndroidKeystoreName: /Users/ncontinanza/key2.keystore
```
tvillegas98, line 258: I think we should not push this change 🤔

client/Assets/Scripts/SocketConnectionManager.cs
```diff
@@ -52,6 +50,16 @@ public class Session
...
     {
+        Init();
+    }
+
+    public void Init()
+    {
+        if (Instance != null)
+        {
+            Destroy(gameObject);
+            return;
```
BertovDev, line 61: This looks great. Just for curiosity why are you using `return` here?

BertovDev (changes requested):

client/Assets/Scenes/BackendPlayground.unity
BertovDev, line 1: Please remember that changes have to be done inside the respective prefab and then applied to the scene.
In this case this changed had to be done inside `CameraUI` prefab.

BertovDev (changes requested):

client/Assets/Scripts/SocketConnectionManager.cs
ncontinanza: Because this case is the one in which the instance is a duplicate (that's why it's not null), so I call Destroy() to make sure that the Singleton has only one such component in the scene. And then return, because there's nothing else to do with that instance.

client/Assets/Scenes/BackendPlayground.unity
ncontinanza: Right, it's fixed now.

BertovDev (approved):
Awesome , working fine!!

BertovDev (approved):

server/lib/dark_worlds_server/engine/runner.ex
```diff
@@ -248,9 +250,18 @@ defmodule DarkWorldsServer.Engine.Runner do
...
     {:ok, game} = Game.disconnect(game_state.game, player_id)
+
     {:noreply, %{gen_server_state | client_game_state: %{game_state | game: game}, current_players: current}}
   end
 
+  def handle_cast(
+        {:disconnect, _player_id},
+        %{game_status: :character_selection} = gen_server_state
+      ) do
+    current = gen_server_state.current_players - 1
```
manucamejo, line 261: Why we not removing the player from the list?

martinacantaro:
Followed the instructions:

- Created several Quick games
- Pausing games and going back to lobby worked fine every time.
- The Network Manager worked as expected.
- Creating lobbies and going back worked, but around 10% of the time I saw this bug where it kicks you back into the lobby and you have to click Back again. I made a video where it also shows when the Network Managers appears and disappears

https://github.com/lambdaclass/curse_of_myrra/assets/6120897/538babd9-3787-48f0-8971-379f43ce0b6a

server/lib/dark_worlds_server/engine/runner.ex
ncontinanza: Should be fixed now

ncontinanza:
> Followed the instructions:
> 
> * Created several Quick games
> * Pausing games and going back to lobby worked fine every time.
> * The Network Manager worked as expected.
> * Creating lobbies and going back worked, but around 10% of the time I saw this bug where it kicks you back into the lobby and you have to click Back again. I made a video where it also shows when the Network Managers appears and disappears
> 
>  bug.mov

That's an issue that we should take care of in another PR, it only happens in the editor. When you use your mouse to click on a button, a second click will be received if in the next scene there is another button under your mouse pointer. You can easily verify this if, after clicking a button, you quickly move the pointer to a side of the screen or somewhere else where you know there won't be a button in the next scene. On the contrary, if you click in a spot where there is a button in the next scene, you'll see the effect that you mentioned.

BertovDev (approved):
- [x] Tested in Unity Desktop
- [x] Tested in Android 11

PR #547, opened 2023-07-12
Hack attacks details

manucamejo:


server/native/gamestate/src/game.rs
```diff
@@ -875,9 +910,11 @@ impl GameState {
                 }
 
                 let mut kill_count = 0;
-                for target_player_id in affected_players {
+
+                // A projectile should attack only one player per tick
+                if affected_players.len() > 0 {
                     let attacked_player =
-                        GameState::get_player_mut(&mut self.players, target_player_id)?;
+                        GameState::get_player_mut(&mut self.players, affected_players[0])?;
```
tvillegas98, line 917: I'm not sure is this is the best option, but we should test it in a playtesting

server/native/gamestate/src/game.rs
```diff
@@ -839,12 +841,45 @@ impl GameState {
                             speed,
                         )
                         .unwrap();
+
+                        let attack_dmg = 2;
```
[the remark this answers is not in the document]
manucamejo: We should test it, because the damage is per tick

client/Assets/Scripts/PlayerMovement.cs
```diff
@@ -372,6 +372,10 @@ private void movePlayer(GameObject player, Player playerUpdate)
             }
         }
 
+        if (playerUpdate.Effects.ContainsKey((ulong)PlayerEffect.Piercing))
```
tvillegas98, line 375: It would be better if h4ck's 4th skill adds another effect that boosts its speed, doesn't make sense that piercing increases your speed

tvillegas98 (approved):

PR #562, opened 2023-07-14
Get commit hash to match client and server versions

AminArria:
- [x] Get commit hash on server/client build
- [x] Show client/server hash in UI during game

- [ ] Reject game websocket connection if hashes do not match (will be done in a follow up if needed)


manucamejo:
image
Server hash: 7 characters
Client hash: 8 charactes
Can we use the same size of hash?

AminArria:
> Can we use the same size of hash?

Yes, 5c2bc60 fixes this and ask for 8 chars explicitly

tvillegas98 (approved):
Nice work ! 😄

manucamejo (approved):

PR #563, opened 2023-07-16
increase board size to 10000 + improve leap

### LINKED ISSUE #574: [REFACTOR] Muflus leap

### LINKED ISSUE #575: [FEATURE] Increase board size
We need more granularity.
The diagonal movement is very slow because of the `round()` function.

manucamejo:
closes #574
closes #575
- Increase board size to 10k x 10k.
- Leap now is an effect and hit at the end.
- Refactor `players_in_range`, now we calculate the distance between the players, instead of looking into a mini-matrix.
- Add `can_move` function, to check if the player is able to move.


tvillegas98 (approved):

tvillegas98:
Nice work @manucamejo ! 🚀

PR #566, opened 2023-07-17
Fix spameable buttons in lobby

### LINKED ISSUE #536: [BUG] Multiple click events received when clicking a button.
## 🐛 Bug Report

This bug only happens when playing the game in the Unity editor.
When using a mouse to click on a back button, a second click will be received if in the next scene there is another button under the mouse pointer. This behaviour can be easily verified if, after clicking a button, the pointer is quickly moved to a side of the screen or somewhere else where there won't be a button in the next scene. On the contrary, if the button is clicked in a spot where there is a button in the next scene, this effect will be observed. 

### Steps to Reproduce

1. Start the game in the Unity editor. 
2. In the lobbies scene, click "New Lobby".
3. Click on the rightmost region of the "Back" button (so that when the scene changes the pointer is over the "New Lobby" button), and keep the pointer still after clicking.

https://github.com/lambdaclass/curse_of_myrra/assets/17294394/b21b1114-9071-4c8d-bfd4-49f7bb225e92

BertovDev:
Fix the issue #536 , disabling the button so players cant spam it

- [x] Tested in Unity Desktop
- [x] Tested in Android

closes #536

### DIFF
--- a/client/Assets/Prefabs/UI/LobbiesButtons.prefab
+++ b/client/Assets/Prefabs/UI/LobbiesButtons.prefab
@@ -403,6 +403,7 @@ Animator:
   m_HasTransformHierarchy: 1
   m_AllowConstantClipSamplingOptimization: 1
   m_KeepAnimatorStateOnDisable: 0
+  m_WriteDefaultValuesOnDisable: 0
 --- !u!114 &1441671513750593463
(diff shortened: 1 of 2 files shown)


ncontinanza (approved):
LGTM!

PR #569, opened 2023-07-18
[Bug fix] Right Joysticks

AngieDutra:
This PR fixes the disabled styles when pressing but not using a joystick in the right panel


tvillegas98 (approved):
LGTM!

BertovDev (approved):

PR #572, opened 2023-07-18
Shrinking Safe Zone

ncontinanza:
Closes #560 
Closes #455 

This PR creates the typical shrinking safe zone of a battle royale game.
At the beginning and for the first 2 minutes of the match, the whole map will be inside the safe zone. After 2 mintues, the zone will start shrinking until it contains the whole map. If a player is outside the safe zone, it will receive damage once every second.

## Steps to test
1. Start a new match.
2. Wait 2 minutes, or edit the `@map_shrink_wait_ms` in runner.ex.
3. Observe how a red rectangle begins to advance on the map. 
4. Try going inside and outside of the safe zone, and notice how your player's health is affected.

PR #577, opened 2023-07-19
Protobuf messages to indicate map shrinking center and radius

AminArria:


AminArria:
Moved changes into #572

Issue #565, opened 2023-07-17
[FEATURE] Allow client to reconnect to match


AminArria:
## 🚀 Feature

Main problem here is we need way to identify the client/user so we can find and show their current game to allow reconnecting to it

Issue #527, opened 2023-07-12
[FEATURE] Players can choose their name


AminArria:
## 🚀 Feature

A player should be able to choose a name for themselves (maybe in main lobbies screen?) and then we will propagate that to the games so they show up instead of the player IDs (not replace, just for UI stuff, internally should be IDs)

jazimp:
![Image](https://github.com/lambdaclass/curse_of_myrra/assets/134973476/6fb856fc-d174-4a8d-9026-cdfafd6eca6e)

Issue #60, opened 2023-04-21
[FEATURE] User sign up/login through OAuth


AminArria:
## 🚀 Feature

- [ ] Setup Google OAuth for sign up and login
- [ ] Limit allowed emails to lambdaclass domain

marco-paulucci:
the younger and prettier version of this ticket is now in #961

PR #585, opened 2023-07-19
Reconnect to current game

AminArria:
Part of #565 

- [x] Module for keeping track of players and their game
- [x] Endpoint to retrieve current game of a player
- [x] [Temp] Local storage using ETS 
- [ ] ~Shared storage for keeping track~
- [x] Identify players using an external unique id (see #60, #527 ) 

Notes:
- Each client generates a unique id which is then persisted and sent to the server, later on this id is used to retrieve the games. The "problem" with this is that a user can't reconnect from another device and a second user in the same device would reconnect to the other user's game

Pending frontend changes:
- [x] Way to do the reconnect when possible (e.g a button that appears). To know if a reconnection is possible `LobbyConnection.Instance.reconnectPossible`
- [x] When reconnecting to a game there is some transient errors, one of them involving showing the safe zone (everything looks red) which I guess is due to the client appearing suddenly and not having any previous state.

Popup implementation could be already implemented in #677 .
### Graphic reference
https://drive.google.com/file/d/1mlZGLdgkmHFr0d4DlNMO83BpxvP2f7SI/view
image


manucamejo (approved):

ncontinanza (approved):
LGTM!

PR #586, opened 2023-07-20
Dagna basic

martinacantaro:
Ability specification:
> While active, D’agna will move at half speed - playing music while doing a funny walk. It deals damage to all nearby enemies. The further the sound travels, the less damage it deals.

How I implemented the backend side:
- D'Agna receives a Status::Slowed that decreases her speed
- It's an area of effect attack that in fact is a smaller AOE (with full damage) nested inside a larger AOE (with half the amount of damage)


marco-paulucci:
> D'Agna receives a Status::Slowed that decreases her speed

I would suggest using negative movement speed instead (whatever Muflus rage uses but with a negative value).

The reason being: slow is a control effect, and in the future there will be mechanics that act upon control effects, but D'agna's basic isn't meant to be one. It would make her immune to slow, or when she is slowed she could use basic on/off and be back at normal speed, etc. So it's better to decouple it

ncontinanza (dismissed):
LGTM!

server/lib/dark_worlds_server/engine/runner.ex
```diff
@@ -83,6 +83,7 @@ defmodule DarkWorldsServer.Engine.Runner do
     Process.send_after(self(), :all_characters_set?, @character_selection_check_ms)
     Process.send_after(self(), :character_selection_time_out, @character_selection_timeout_ms)
     Process.send_after(self(), :check_player_amount, @player_check)
+    Process.send_after(self(), :shrink_map, @map_shrink_wait_ms)
```
tvillegas98, line 86: Remove this

mgrunwaldt (dismissed):
I left some comments that should be addressed but are not specific from this PR. I'll create refactor issues for them. Before merging lets do a playtest with many players and different characters, I think this PR affects many aspects of the code and we might be missing something out by only looking at the code

mgrunwaldt (dismissed):

mgrunwaldt (approved):

tvillegas98 (approved):

PR #590, opened 2023-07-20
Skill Area Feedback

### LINKED ISSUE #418: [FEATURE] Skill area of effect feedback
When we execute an skill we are triggering the animation but we don't have a proper feedback of the area of effect of it. We should add one so that players have that information.

Let's start with a simple generic one and then add the ability to customize it in a future task.

Waiting: https://github.com/lambdaclass/curse_of_myrra/issues/431

BertovDev:
Implements skill area feedback to know the max range of it. For now the values are hardcored since we dont have the correct config for it in the skill.json. 
You can change the radius in the respective scripteableObject of the skill changing the `SkillAreaRadius`

To test it:

1. Pick 🐂Muflus🐂 and use roll
2. Stop the game, change `SkillAreaRadius` in the `Muflus_roll` scripteableObject
3. Use roll again

closes #418


ncontinanza (dismissed):
Nice job! Tested both in the editor and with an Android build, works like a charm 🚀

ncontinanza (dismissed):

marco-paulucci:
Hi! I believe the area parameter was implemented in the JSON files.

ncontinanza (approved):

manucamejo (approved):

PR #591, opened 2023-07-20
Remove board grid structure

### LINKED ISSUE #507: [REFACTOR] Look into removing/disabling game grid
## 🚧 Refactor

It looks like we are not really using the game grid for anything, most cases we are using the list of entities to run the game. So let's look into removing  it as not having it would probably improve performance and simplify the code

Also, removing the grid will allow us to keep old game states and apply lag compensation logic if needed

manucamejo:
- This PR removes grid structure because it spends a lot of ram and closes #507 

To test it, play it, you should play normally


AminArria (approved):
LGTM

tvillegas98 (approved):

PR #599, opened 2023-07-21
Uma skills

manucamejo:


server/lib/dark_worlds_server/engine/runner.ex
```diff
@@ -82,6 +82,7 @@ defmodule DarkWorldsServer.Engine.Runner do
     Process.send_after(self(), :all_characters_set?, @character_selection_check_ms)
     Process.send_after(self(), :character_selection_time_out, @character_selection_timeout_ms)
     Process.send_after(self(), :check_player_amount, @player_check)
+    Process.send_after(self(), :shrink_map, @map_shrink_wait_ms)
```
tvillegas98, line 85: Remove this

tvillegas98 (changes requested):

server/lib/dark_worlds_server/engine/runner.ex
manucamejo: Done

ncontinanza (changes requested):
There's an overlapping of the poison ability with the mark, in a way that the damage due to the poison is never being applied.

tvillegas98 (approved):
LGTM!

PR #611, opened 2023-07-24
Character selection UI-UX

AngieDutra:
This PR changes the character selection display and styles
It should work as before, however I added a couple new features because of the new additions:
- Added all the players (except from D'Agna to avoid mayor conflicts) and the ones that are not ready to be selected should appear with a lock and blur effect.
- There should be a selection style on each selectable player (and the ones that aren't shouldn't be able to be clicked on)
- You should be able to see each ability and its description according to your selection by a click
- Your name should be the only violet text in the players list
- Solves issue #621 

To do in another PR:
- Adjust D'Agna to the new changes made here


BertovDev (approved):
🤩 Amazing work! 🤩

saugon (approved):
Nice :)

PR #616, opened 2023-07-24
Poisoned Feedback

### LINKED ISSUE #600: [FEATURE] Poisoned feedback
We should create a poisoned feedback to display when a character is poisoned.

ncontinanza:
Closes #600 
This PR implements a feedback for the poisoned status in the three currently available characters.

How to test:
1. Start a match, where at least one of the playing characters is Uma.
2. Go near to a rival and use the poison skill.
3. You should notice a cloud of green smoke that represents the poisoned status. It should go off after 5 seconds.


saugon (approved):
Nice!

PR #619, opened 2023-07-24
Lobby Music

### LINKED ISSUE #540: [FEATURE] Add music to Lobby phase
We could use this song:
https://drive.google.com/file/d/18Ii8k_758oXd-TvpWkUoMQmYXisyAEyj/view?usp=drive_link

Or this one, but only if we use a different one for the Otobi Arena: https://drive.google.com/file/d/1yjl2Q1xec92yOscwgfWMniPLmxeJehYS/view?usp=drive_link

ncontinanza:
Closes #540 
This PR adds background music to the lobby and character selection scenes.

Things to test:
- When you start the application, you shouldn't hear any music. You can start it by increasing the slider placed on the upper right-hand corner.
- Whether you advance to the new lobby or the quick game scenes, you should still hear the music. 
- In every scene, you should be able to control the music volume with the slider. This volume level should remain the same when you advance to the next scene.
- When the match begins, the lobbies background music should stop and you shouldn't be hearing any music nor sound effects.
- If you unmute the match's music and sfx in the pause screen, everything should still be working as usual.

There are a few behaviours that we know are present and they will be fixed in the near future:
- Whenever you refresh or reload the lobbies scene, the music mutes and restarts. (#634)
- When you toggle the music during a match, the volume is always set to 1. (#705)


tvillegas98:
I found a bug where if you didn't turn on the sound in lobbies/lobby, in game turning on the sound doesn't work 🤔

BertovDev (changes requested):
Good work!! But the refresh button in the lobbies scene is breaking everything! You can create ilimitate instances of the MMAudioSourcePool and no one is actually playing the music

BertovDev (approved):
Look great! 🤩

ncontinanza:
> I found a bug where if you didn't turn on the sound in lobbies/lobby, and then if you try to turn on the sound in game, doesn't reproduce any sound 🤔

Nice catch! Should be fixed now.

BertovDev (approved):

saugon (approved):

PR #647, opened 2023-07-26
World tick refactor

fkrause98:
## Description
The World Tick function was getting too long and clumsy, the intent of this PR is to
make it more modular and descriptive.
## Changes
- Introduced new type MutablePlayer.
   - I'm expecting to eventually turn this into the
  actual Player type. 
  - The rationale behind this change is that a lot of times we have
  interactions between 2 (or more) players and a lot can happen. For example, let's say player 1 attacks 
  player 2, this case is easy, we need only to mutate player 2 and we're done. But what when
  also want to register points for player1? What if player 1 receives a health boost after 
  attacking player 2?
  - Now we need them both to be mutable, but we can't do that with
  Rust's borrow checker, so that's why I've decided to create Mutable player which
  simply wraps up the current Player type with an [Rc<RefCell<T>>](https://rustwiki.org/en/book/ch15-05-interior-mutability.html).
- Introduced TickState, since a lot can happen in a Tick, it makes sense to have a new
  struct that holds this state change.
- GameSate::world_tick is now separated into several, and hopefully, self-describing functions.


Oppen:
Maybe using a struct of vectors instead of vector of structs could help here. We could keep a hit-point delta vec (so we update hit-points for each player at the end of the tick), another for points, etc. With this you could just have the mutable vector or slice and update in place the single number.

server/native/gamestate/src/game.rs
```diff
@@ -1093,202 +1080,101 @@ impl GameState {
...
+            &dashing_player_position,
+            direction,
+            dashing_player_speed as i64,
+        )?;
+        player.set_position(new_position);
+        *affected_players = GameState::affected_players(
+            2,
+            200.,
+            // TODO: Remove this if we ever use MutablePlayers
+            (players.clone()).into_iter().map(Into::into).collect(),
```
Oppen, line 1105: I'm not sure you need to clone the vector here. You could do this instead:
```suggestion
            players.iter().map(Into::into).collect(),
```
Or if that doesn't work:
```suggestion
            players.iter().cloned().map(Into::into).collect(),
```

fkrause98: Hey @Oppen 
Thanks for the comment! I'll try that and let you know 👍

server/native/gamestate/src/tick_changes.rs
```diff
@@ -0,0 +1,377 @@
...
+        Ok(())
+    }
+    pub fn accumulate_dashing_effects(
+        &mut self,
+        player: &MutablePlayer,
+        expired_effects: &[(Effect, EffectData)],
+        players: &MutablePlayers,
+        board: &Board,
+    ) -> Result<(), String> {
+        match player.character().name {
```
AminArria, line 310: Why check for character name? If you only check for the effects you want (e.g. `NeonCrashing`, `Leaping` etc) it would be enough and makes changing this later on easier

AminArria (approved):
LGTM, just a minor improvement on the comment I left

server/native/gamestate/src/tick_changes.rs
fkrause98: That's a good idea, I'll try some short changes to see how this turns out and let you know.

PR #651, opened 2023-07-26
Extend Healthbar graphic

BertovDev:
This pr extend the health bar graphic by 25% , also adds missing poison prefab to the BOT prefab.


ncontinanza:
I noticed that if you play with Uma, the health bar and the player's name overlap. I checked in main and the problem is not from your PR, but do you think it would be easy to fix it as a part of this feature? Or should we leave it for another issue?

image

BertovDev:
> I noticed that if you play with Uma, the health bar and the player's name overlap. I checked in main and the problem is not from your PR, but do you think it would be easy to fix it as a part of this feature? Or should we leave it for another issue?
> image

This is pretty easy to fix, it happends because Uma model scale is wrong.

manucamejo (approved):

tvillegas98 (approved):

PR #691, opened 2023-07-31
Config screen

lotuuu:
Adds a screen to edit the json config files. Implemented config files are `Characters.json`, `Skills.json` and `GameSettings.json`. Implemented in liveview to handle the addition of new entries.

Characters and skills are listed by name. Name (and id for Characters) are required fields. For Game Settings, all fields are required.

Addition of new characters and skills, as well as custom values for them, has been implemented. Values need a name to be created, while entities don't (they get assigned an indexed name).

Screen should be styled later on.


server/lib/dark_worlds_server/utils/config/config.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Utils.Config do
+  @streaming_assets_path "../client/Assets/StreamingAssets/"
```
AminArria, line 2: Not to solve here, but better mention it

This will break once we split the repos

tvillegas98 (dismissed):
Nice work! 😄

server/lib/dark_worlds_server/utils/config/config.ex
lotuuu: Yes, this will be fixed in a future PR!

BertovDev (approved):
Superb !! 🤩

Issue #687, opened 2023-07-28
Killfeed doesn't count kills correctly


manucamejo:

mztrpln:
We've tested today twice with games with over 10 players and the killfeed count was correct both times.

PR #693, opened 2023-07-31
fix kill count

### LINKED ISSUE #571: [BUG] Kill messages appears although the player is dead

manucamejo:
Closes #571 #687 

To test it, choose any character, start a match against another player, and kill each other. Both the killfeed and the number of kills should work correctly.


tvillegas98 (approved):
LGTM!

PR #696, opened 2023-08-01
Fix re-entry in a matchmaking lobby not removing players

### LINKED ISSUE #326: [FEATURE] Disconnect the player from the game when they click back to lobby
## 🚀 Feature

In the server we have a base logic to handle disconnects from players, but we're not sending a message to the server from the client if a player disconnected via button "Back to lobby"

AminArria:
Closes #326 

- [x] Close web socket connection when instantiating a new `LobbyConnection` (a.k.a joining a lobby)
- [x] Next `player_id` is calculated in a smarter way, we get the list of current ones and either generate the next one or use the first one missing
- [x] When joining a lobby wait to have a player id to render the scene
- [x] Destroy lobby if empty 
- [x] Select a new "host" when current one disconnects
- [x] Decouple player_id and player_name


manucamejo (dismissed):

tvillegas98 (approved):
Nice work ! 🪨

manucamejo (approved):

PR #698, opened 2023-08-01
Fix use of interpolation for your own character when not using client prediction

jrchatruc:
Currently, if you disable client prediction interpolation does not kick in for your character, even if enabled. This fixes that


jrchatruc:
This merged as part of #678

PR #699, opened 2023-08-01
Fix interpolation breaking when alt tabbing

jrchatruc:
There was a bug on interpolation where if you moved the game to the background, the interpolation delay would be set to the max when going back into the game. This was due to the use of `Time.deltaTime` to keep track of accumulated time. 

The problem is that if the game is not running, this game time does not happen and therefore the accumulated time lags behind. 

The fix was to switch to using unix timetamps to keep track of time.

TODO:

- [x] Document this


jrchatruc:
This was merged as a part of #678

PR #703, opened 2023-08-01
Fixed unwanted Muflus's leap feedback loop

saugon:
Ditto.


mgrunwaldt (approved):
Tested in Unity simulator + iOS, worked perfectly in both cases (before in iOS muflus animation after leap repeated). Didn't test Android

BertovDev (approved):
- [x] Tested in Unity Desktop
- [x] Tested in Android

PR #712, opened 2023-08-02
Small fix to interpolation

jrchatruc:
When getting the previous and next events to check for movement, we were using time (specifically, a hardcoded 30 ms) to try to index into the next and previous elements of the buffer. This can be a problem because updates are not guaranteed to always be 30 ms apart, for a variety of reasons.

Instead, this PR changes things so we simply get the next and previous events with their index. Events are guaranteed to be ordered so this is enough


tvillegas98 (approved):
Nice catch! thanks 😄

PR #720, opened 2023-08-03
Hack ultimate feedback

AngieDutra:
This PR gives feedback when the hack ultimate skill is activated
This skill should have 6 seconds of duration. Check it is working for everyone playing and that the time duration is okey.


saugon (approved):
Great job! ⚡

PR #726, opened 2023-08-03
Network error modal

### LINKED ISSUE #677: [FEATURE] Errors
## 🚀 Feature

Implement an error dialog so that the client can see when something goes wrong. Some examples:
- Disconnect from game
- Fail to join lobby
- Fail to connect to server 

### Graphic reference
https://drive.google.com/file/d/1ZVE93nwd_-3ti41tXj7enSizKwoW3hFB/view?usp=sharing
image

AminArria:
Closes #677


client/Assets/Scripts/LobbyConnection.cs
```diff
@@ -207,6 +212,7 @@ public IEnumerator StartGame()
 
     public void Reconnect()
     {
+        Debug.Log("reconect");
```
ncontinanza, line 215: Should we remove this log? 
In case it's necessary to have it there, it contains a typo ("reconnect")

client/Assets/Scripts/SocketConnectionManager.cs
```diff
@@ -205,6 +206,15 @@ private void OnWebSocketMessage(byte[] data)
         }
     }
 
+    private void onWebsocketClose(WebSocketCloseCode closeCode)
```
ncontinanza, line 209: Following the C# naming convention, this should be `OnWebSocketClose(...)`

client/Assets/Scripts/LobbyConnection.cs
```diff
@@ -412,6 +417,15 @@ private void OnWebSocketMessage(byte[] data)
         }
     }
 
+    private void onWebsocketClose(WebSocketCloseCode closeCode)
```
ncontinanza, line 420: Following the C# naming convention, this should be OnWebSocketClose(...)

ncontinanza (dismissed):
Nice job! I left a few minor comments about the code.

Also, I noticed that when I kill the connection to the server in order to test the new functionality, the buttons I use don't go back to their "not-pressed" state. Is this something that this PR should consider?

image

AngieDutra:
> Nice job! I left a few minor comments about the code.
> 
> Also, I noticed that when I kill the connection to the server in order to test the new functionality, the buttons I use don't go back to their "not-pressed" state. Is this something that this PR should consider?
> 
> image

This happens due to the button selection script. I think it is not detecting a pointer up when the pop up is activated, so it stays with pressed styles until refresh. But I should look into all buttons and how they are working because I've found several issues regarding this. Also, what I change in this script will affect all buttons so I would do it in another PR. You could do an issue for this tho

client/Assets/Scripts/LobbyConnection.cs
AminArria: Removed

client/Assets/Scripts/LobbyConnection.cs
AminArria: Renamed

client/Assets/Scripts/SocketConnectionManager.cs
AminArria: Renamed

BertovDev:
> Nice work @AminArria and @BertovDev! I enjoyed working together

Same for me!!!

BertovDev (approved):
Awesome!! 🤩

PR #735, opened 2023-08-04
Fix missing prefab in lobbyManager

BertovDev:
This Fix the console errors where playbutton.

PD : We should refactor the use of lobbyManager removing it from the pause splash , chatacter selection.


AngieDutra (approved):
Nice! ✨

saugon (approved):

PR #751, opened 2023-08-07
Basic code refactor / clean up

saugon:
- Renamed scene: BackendPlayground -> Battle.
- Renamed script: Playermovement.cs -> Battle.cs.
- Improved Battle.cs code readability and removed unused code/attributes.
- Removed unused objects/script from MMT.
- Converted managers object into prefabs (except sound manager, which is being refactored in another issue).


mgrunwaldt (approved):
Tested in iOS and unity simulator

saugon:
> Nice job! The code seems much cleaner now. A future step could be the task of separating the Battle class into smaller classes with less responsibility.

Totally agree 👍🏻

PR #755, opened 2023-08-08
Start documentation for entity interpolation

### LINKED ISSUE #746: [DOCS] - Finish fluid gameplay documentation
## 📚 Documentation

- [ ] Entity interpolation
- [ ] Action rate, hacky solution

tvillegas98:
This PR adds docs for Entity Interpolation in the Fluid Gameplay section

Closes #746


docs/src/fluid_gameplay.md
```diff
@@ -205,10 +205,49 @@ To avoid both these things, the speed used by the client needs to exactly match
...
-TODO
+- We have two players in the game (Player 1 and 2)
+
+Our movement feels pretty smooth now, but what about Player 2?. His movement only updates every time we receive an update from the server and supposing we receive an update every 20ms, we are receiving 50 game updates times per second.
+
+Now, let's see Player 2's movement:
+
+![](./videos/Interpolation_at_0ms.gif)
+
+As you can see, his movement feels weird sometimes it suddenly stops and that's not because Player 2 decided to stop, it is because we're updating their movement based what on the updates sent by the server. How can we solve this?. Well, here's where Entity Interpolation comes to the battleground. From now on, we'll be calling it just *Interpolation*.
```
manucamejo, line 219: "based what on the updates sent by the server." remove 'what'

docs/src/fluid_gameplay.md
```diff
@@ -205,10 +205,49 @@ To avoid both these things, the speed used by the client needs to exactly match
...
+
+Theory behing Interpolation is that we render an X amount of time in the past what other player are doing, and that allows us to interpolate the movement. Let's suppose we're interpolating 100ms in the past, player's movement would like this
```
manucamejo, line 221: "behing" -> behind

docs/src/fluid_gameplay.md
```diff
@@ -205,10 +205,49 @@ To avoid both these things, the speed used by the client needs to exactly match
...
+
+![](./videos/Interpolation_at_100ms.gif)
+
+Looks smoother now, doesn't it? Well, let's increase that interpolation time and turn on the server's ghost, this will allows us to see what exactly is Interpolation attacking
```
manucamejo, line 225: "allows" -> allow

docs/src/fluid_gameplay.md
mgrunwaldt: I would put "our *player's* movement feels.."

docs/src/fluid_gameplay.md
mgrunwaldt: You are talking about "interpolate the movement", I don't know if people who never worked with this in the past will understand what this mean. Later on it's explained, but here you use the word before actually explaining what it means, I would try to go one step back and write it for someone who doesn't know anything about this subject before even using the word.

mgrunwaldt (changes requested):
Nice work, I left a small comment

docs/src/fluid_gameplay.md
tvillegas98: Yes. you're right, i'll be changing this to be more friendly to newcomers

docs/src/fluid_gameplay.md
tvillegas98: Done!

docs/src/fluid_gameplay.md
tvillegas98: Done!

docs/src/fluid_gameplay.md
tvillegas98: Done!

mgrunwaldt (approved):

AngieDutra (changes requested):

docs/src/fluid_gameplay.md
```diff
@@ -209,23 +209,23 @@ For this topic, lets set up our scenario:
...
+Our movement feels pretty smooth now, but what about Player 2?. His movement only updates every time we receive an update from the server and supposing we receive an update every 20ms, we are receiving 50 game updates per second.
 
 Now, let's see Player 2's movement:
 
 ![](./videos/Interpolation_at_0ms.gif)
 
-As you can see, our player's movement feels that sometimes it suddenly stops and that's not because Player 2 decided to stop, it is because we're updating their movement based on the updates sent by the server. How can we solve this?. Well, here's where Entity Interpolation comes to the battleground. From now on, we'll be calling it just *Interpolation*.
+As you can see, his movement feels that sometimes it suddenly stops and that's not because Player 2 decided to stop, it is because we're updating their movement based on the updates sent by the server. How can we solve this? Well, here's where Entity Interpolation comes to the battleground. From now on, we'll be calling it just *Interpolation*.
```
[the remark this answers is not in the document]
mgrunwaldt: Not sure I agree on this, if I read our player's movement I will automatically think on my player, not on player 2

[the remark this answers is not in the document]
AngieDutra: player 2's could be an alternative but if not you could leave it as it is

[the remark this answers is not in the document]
mgrunwaldt: lets go for player 2's

AngieDutra (dismissed):
LGTM! 💥

docs/src/fluid_gameplay.md
```diff
@@ -2,9 +2,9 @@
...
 
-Below we go over the problems that arise from networking, how we solved some of them, and how we might improve these solutions in the future. This section is going to be long, full of explanations and videos showing different issues that arise from playing through a network. I HIGHLY encourage anyone reading this to actually try this stuff out by themselves. The way we judge whether we have done a good job or not is if the game feels good to play. Ultimately, that is the only metric that matters here.
+Here we'll address the problems that arise from networking, how we solved some of them, and how we might improve these solutions in the future. I HIGHLY encourage anyone reading this to actually try this stuff out by themselves. The way we judge whether we have done a good job or not is if the game feels good to play. Ultimately, that's the only metric that matters here.
```
ncontinanza, line 5: This part starts in the plural first-person, then changes to singular first-person and then goes back to the plural first-person. Maybe we could change the second sentence to "We HIGHLY encourage anyone..."

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
 To avoid both these things, the speed used by the client needs to exactly match the one used by the server.
 
 ## Entity interpolation
+We have made movement look smooth when playing over `localhost`. When moving to actual remote servers, however, our code so far is still not robust enough. There's still one thing to do, and that is *Entity Interpolation*.
+For this topic, lets set up our scenario:
```
ncontinanza, line 208:
```suggestion
For this topic, let's set up our scenario:
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
 
-We have made movement look smooth when playing over `localhost`. When moving to actual remote servers, however, our code so far is still not robust enough.
+- We have two players in the game (Player 1 and 2)
 
-TODO
+Our movement feels pretty smooth now, but what about Player 2?. His movement only updates every time we receive an update from the server and supposing we receive an update every 20ms, we are receiving 50 game updates per second.
```
ncontinanza, line 212:
```suggestion
Our movement feels pretty smooth now, but what about Player 2?. His movement only updates every time we receive an update from the server and, supposing we receive an update every 20ms, we are receiving 50 game updates per second.
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
+
+Now, let's see Player 2's movement:
+
+![](./videos/Interpolation_at_0ms.gif)
+
+As you can see, Player 2's movement feels that sometimes it suddenly stops and that's not because Player 2 decided to stop, it is because we're updating their movement based on the updates sent by the server. How can we solve this? Well, here's where Entity Interpolation comes to the battleground. From now on, we'll be calling it just *Interpolation*.
+
+The theory behind Interpolation is that we save the events returned by our server, allowing us to render movements that have already happened. In addition, we are able to predict the movement of the following events. Let's suppose we're interpolating 100ms in the past, a player's movement would look like this
```
ncontinanza, line 220:
```suggestion
The theory behind Interpolation is that we save the events returned by our server, allowing us to render movements that have already happened. In addition, we are able to predict the movement of the following events. Let's suppose we're interpolating 100ms in the past, a player's movement would look like this:
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
+
+![](./videos/Interpolation_at_100ms.gif)
+
+Looks smoother now, doesn't it? Well, let's increase that interpolation time and turn on the server's ghost, this will allow us to see what exactly Interpolation is doing
```
ncontinanza, line 224:
```suggestion
Looks smoother now, doesn't it? Well, let's increase that interpolation time and turn on the server's ghost, this will allow us to see what exactly Interpolation is doing.
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
+
+![](./videos/Interpolation_with_ghost.gif)
+
+The Player 2 that's behind is the one you're really seeing when you're playing, and the one that's in front is showing the latest update you received from the server. As we can see, this looks choppy and sometimes jitters.
+
+How did we achieve this?. Basically we're doing a few things:
+1. We keep track of time that has passed since the first update
```
ncontinanza, line 231:
```suggestion
1. We keep track of the time that has passed since the first update.
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
+2. We store the last 20 updates sent by the server
```
ncontinanza, line 232:
```suggestion
2. We store the last 20 updates sent by the server.
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
+3. We're *Interpolating* the movement of other players by checking what they did in a previous update, what they're doing in the current update and what he will do in the next update. 
+
+All of this code is in the `EventsBuffer.cs`. Where you can change the delta interpolation time as you please. Remember, the delta interpolation time is how much time in the past you will render, if it's too high, you'll lying too much to your players.
```
ncontinanza, line 235:
```suggestion
All of this code is in the `EventsBuffer.cs`, where you can change the delta interpolation time as you please. Remember, the delta interpolation time is how much time in the past you will render, if it's too high, you'll be lying too much to your players.
```

docs/src/fluid_gameplay.md
```diff
@@ -204,11 +204,48 @@ If we go too fast, we will stop in our tracks for a little while, waiting for th
...
 
 ## Action rate, hacky solution
 
-TODO
+As we said in the previous topic, we keep track of the accumulated time since we received the first update from the server to interpolate the movement. The theory behind this should work, but we found that there were times where the interpolation ghosts differed too way much from the server. After a bit of research, we discovered that the problem was with using the [`Time.deltaTime`](https://docs.unity3d.com/ScriptReference/Time-deltaTime.html) that belongs to Unity. That amount of time is how much time has passed since the last frame to the current one, that sounds okay but the problem is that the game editor/android build/ios build should be active in the screen, if you change tabs or let the game run in the background the `Time.DeltaTime` stops but the server doesn't and because of that all the operations get messed up. In order to solve this, we just change it for Unix time in milliseconds.
```
ncontinanza, line 239:
```suggestion
As we said in the previous topic, we keep track of the accumulated time since we received the first update from the server to interpolate the movement. The theory behind this should work, but we found that there were times where the interpolation ghosts differed too way much from the server. After a bit of research, we discovered that the problem had to do with using the [`Time.deltaTime`](https://docs.unity3d.com/ScriptReference/Time-deltaTime.html) that belongs to Unity. That amount of time is how much time has passed since the last frame to the current one. That sounds okay, but the problem is that the game editor/Android build/iOS build should be active in the screen, if you change tabs or let the game run in the background the `Time.DeltaTime` stops but the server doesn't and because of that all the operations get messed up. In order to solve this, we just changed it for Unix time in milliseconds.
```

ncontinanza (changes requested):

docs/src/fluid_gameplay.md
```diff
@@ -27,11 +29,11 @@ while true {
...
+This mental model is so common in games that it has a name; this is the `game loop`. A `frame` is the result of one iteration of this loop. In multiplayer games, such as ours, this loop looks slightly different: clients are not the ones updating game state, they get those updates from the server. The general idea still applies however.
  
-Typically, the more computing power you have, the higher your framerate. In Myrra, we show players their framerate in the bottom left corner of the screen. Its value is usually capped at `300`, and anything below `30` will feel really bad to play. Most games run either at `30`, `60` or at an uncapped `FPS`.
+Typically, the more computing power you have, the higher your framerate. In Myrra, we show players their framerate in the bottom left corner of the screen. Its value is usually capped at `300`, and anything below `30` will feel really bad to play. Most games run either at `30`, `60` or at an uncapped `FPS` rate.
```
ncontinanza, line 34:
```suggestion
Typically, the more computing power you have, the higher your framerate. In Myrra, we show players their framerate in the bottom of the screen. Its value is usually capped at `300`, and anything below `30` will feel really bad to play. Most games run either at `30`, `60` or at an uncapped `FPS` rate.
```

ncontinanza: This was outdated, we changed that design a few weeks ago.

docs/src/fluid_gameplay.md
```diff
@@ -45,53 +47,67 @@ The `Tick Rate` is the rate at which the server sends game updates to every clie
...
-- The second one is the result of network instability. Servers send game updates at a fixed frequency (the `tick rate`), and we assume clients get them at that same frequency. This assumption is _wrong_. Sometimes network packets/frames take longer than expected, sometimes they get lost and have to be retransmitted (we use `Weboscket`, which uses `tcp` underneath). This means clients may not get any update from the server for a little while, only to then get a few of them all at once.
+- The second one is the result of network instability. Servers send game updates at a fixed frequency (the `tick rate`), but it would be wrong to assume clients get them at that same frequency. Sometimes network packets/frames take longer than expected, sometimes they get lost and have to be retransmitted (we use `Weboscket`, which uses `tcp` underneath). This means clients may not get any update from the server for a little while, only to then get a few of them all at once.
```
ncontinanza, line 102:
```suggestion
- The second one is the result of network instability. Servers send game updates at a fixed frequency (the `tick rate`), but it would be wrong to assume clients get them at that same frequency. Sometimes network packets/frames take longer than expected, sometimes they get lost and have to be retransmitted (we use `Websocket`, which uses `TCP` underneath). This means clients may not get any update from the server for a little while, only to then get a few of them all at once.
```

docs/src/fluid_gameplay.md
```diff
@@ -110,7 +126,7 @@ Input lag was the first issue we tackled, even though in hindsight it's clearly
...
 
-Notice that, by doing this, we are lying to the player. When they move in a direction, they immediately see themselves forward, but that's not where they actually are; the server still needs to process that command. The higher the ping, the more we lie. This is a tradeoff. The game feels better to play with prediction on, but lie to players too much and they might notice. They may see shots that look like they should miss hit them, because they are not seeing their character where it actually is.
+Notice that, by doing this, we are "lying" to the player. When they move in a direction, they immediately see themselves forward, but that's not where they actually are; the server, which is the source of truth of the gamestate, still needs to process that comman and send it out to every player. The higher the ping, the further the gamestate will differ from what the player sees on their screen. This is a tradeoff. The game feels better to play with prediction on, but lie to players too much and they might notice. They may see shots that look like they should miss hit them, because they are not seeing their character where it actually is.
```
ncontinanza, line 129:
```suggestion
Notice that, by doing this, we are "lying" to the player. When they move in a direction, they immediately see themselves forward, but that's not where they actually are; the server, which is the source of truth of the gamestate, still needs to process that command and send it out to every player. The higher the ping, the further the gamestate will differ from what the player sees on their screen. This is a tradeoff. The game feels better to play with prediction on, but lie to players too much and they might notice. They may see shots that look like they should miss hit them, because they are not seeing their character where it actually is.
```

docs/src/fluid_gameplay.md
```diff
@@ -179,7 +195,7 @@ More precisely, what we do to fix the jitter above for our character is the foll
...
 
-In other words, what we are doing is making characters chase the server's latest position at their corresponding speed, showing movement gradually over the course of however many frames happen between each tick. When we get a new updated position from the server, instead of immediately placing them there, we move characters smoothly from where they are up to where they need to be. This process of taking two discrete points and smoothing out the movement between them is sometimes called `interpolation`, `linear` `interpolation` or `lerp` for short.
+In other words, what we are doing is making evey character chase the server's latest position at their corresponding speed, showing movement gradually over the course of however many frames happen between each tick. When we get a new updated position from the server, instead of immediately placing them there, we move characters smoothly from where they are up to where they need to be. This process of taking two discrete points and smoothing out the movement between them is sometimes called `interpolation`, `linear interpolation` or `lerp` for short.
```
ncontinanza, line 198:
```suggestion
In other words, what we are doing is making every character chase the server's latest position at their corresponding speed, showing movement gradually over the course of however many frames happen between each tick. When we get a new updated position from the server, instead of immediately placing them there, we move characters smoothly from where they are up to where they need to be. This process of taking two discrete points and smoothing out the movement between them is sometimes called `interpolation`, `linear interpolation` or `lerp` for short.
```

ncontinanza (changes requested):

martinacantaro:
Thanks for your detailed comments @ncontinanza ! They are all resolved now.

ncontinanza (approved):
LGTM! 🚀

mgrunwaldt (approved):

PR #762, opened 2023-08-08
Sound Manager Refactor

### LINKED ISSUE #634: [REFACTOR] Do not restart music when going back in the lobbies scenes
## 🚧 Refactor

Currently, if you move forward in the scenes previous to starting a match (lobbies, lobby, character selection), the music is not affected and keeps playing. But if you go back to a previous scene, as the LobbyConnection is destroyed and a new instance is created, the music starts over and is muted. It would be desirable to continue playing the music as in the case of moving to the next scene.

### LINKED ISSUE #705: [BUG] When music is toggled during a match, volume is always set to 1
## 🐛 Bug Report

After starting a new match, independently of the value set on the music volume slider, the toggle button logic sets the volume in 1.

### LINKED ISSUE #760: [REFACTOR] Sound manager
## 🚧 Refactor

The current implementation of the sound manager it's working but has some know issues that should be fixed.
It may also be better to refactor it from scratch.

### Know issues
#705 
#634 
#561

ncontinanza:
Closes #760 
Closes #634 
Closes #705 

This PR implements a refactor of the way we use the MMSoundManager class. Instead of manually handling MMF_Player objects, we'll now use the BackgroundMusic and PersistentBackgroundMusic classes that TopDown Engine provides. Also, improvements were made in the implementation of the volume slider and the toggle button components.

Things to test:
- The toggle button works correctly in every scene.
- The volume slider correctly controls the volume of the background music during a match.
- The lobbies background music does not restart when using the refresh button nor when changing from one server to another.
- If you turn on the music in the lobbies scene, the match starts with the music on. If the music is off before starting the match, it does not turn on when the match starts.


BertovDev (dismissed):
Superb!!!! 🤩

BertovDev (approved):
Awesome job!

tvillegas98 (approved):
Nice work @ncontinanza ! Everything's working fine now 😄

PR #763, opened 2023-08-08
Staging environment automatic deployment

samoht9277:
### [WIP] Staging environment

The purpose of this new environment is to only deploy _stable_ versions. For other purposes, please use the [testing](https://github.com/lambdaclass/curse_of_myrra/deployments/testing) environment.


klaus993 (approved):
🚀

Issue #643, opened 2023-07-26
[FEATURE] H4ck Mechanical Pass


marco-paulucci:
## 🚀 Feature

This issue aims to list the missing changes to complete H4ck mechanically.

#### Slingshot 
- [x] Parametrize projectile speed in Par1 (fill Par1Desc too)
- [ ] Allow holding button to see a targetting cone like Multishot
- [ ] If there is a target in the cone, it will auto-aim like it works now

#### Multishot ✅ 
- [x] #771

#### Disarm
- [ ] It needs a feedback. Probably some **subtle** body or ground effect (could look like tiny electric pulses).
- [ ] It should use a different ButtonType. It's using Directional rather than Targeted (should work like Muflus' Leap).
- [x] #765
- [x] Rename the skill to Paralyze

#### Denial of Service ✅ 
- [x] Multishot should pierce as well
- [x] H4ck's speed shouldn't increase
- [x] Projectile speed for Sling/Multishot should increase by 25%
- [x] #769
- [x] #445

marco-paulucci:
see #794 794

PR #766, opened 2023-08-09
Disarm to Paralize

### LINKED ISSUE #765: Change behavior. Add a new Status "Paralyzed" in which you can use skills but not move (opposite of disarm). We can keep current disarm for a future character.

manucamejo:
Closes partially #643 
Closes #765 

To test it, choose hack, and use the 2nd skill, the enemy must be paralyzed, and not disarmed.
Now every time a player gets paralyzed, if client prediction is on, we will correct their position to the server's position


BertovDev (approved):
Awesome! 🤙

PR #767, opened 2023-08-09
Add radius for bullets

tvillegas98:
This PR changes the logic for projectiles:
- Now projectiles have a body size
- Reworking the former collision system, that uses intersections between lines, for intersections between circles

### DIFF
--- a/server/native/gamestate/src/game.rs
+++ b/server/native/gamestate/src/game.rs
@@ -317,45 +317,55 @@ impl GameState {
             })
             .for_each(|player| {
                 let radius = player.character.body_size;
+                let projectile_radius = 10.;
 
-                let player_attacked = match p2.y == p1.y {
-                    true => {
-                        // The projectile is moving vertically
-                        let intersection = Position::new(player.position.x, p1.y);
-
-                        intersection.x >= p1.x && intersection.x <= p2.x && // The player is in the projectile's segment
-                        (distance_to_center(&player, &intersection) <= radius) // The player is near the intersection
-                    }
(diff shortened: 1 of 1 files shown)


marco-paulucci (commented):
> Now projectiles explodes in player's current position

this doesn't seem to look proper. I recommend using the client to determine location for missile explosion animation. Also it probably should not be centered to the char but rather appear in the intersection where the collision occurred

server/native/gamestate/src/game.rs
```diff
@@ -296,61 +296,30 @@ impl GameState {
...
+                // Sum the character's radius body size and the projectile radius,
+                // Calculate the distance between the centers of both circles (character's body size center and projectile position)
+                // And then we check if the sum of the radius is greater than the distance between
+                // centers, then the projectile hitted the target :)
+                let character_body_size = player.character.body_size;
+                let projectile_radius = 10.;
+
+                let centers_distance =
+                    distance_between_positions(&projectile_position, &player.position);
+                let radiuses_sum = character_body_size + projectile_radius;
```
Oppen, line 321: Is the body size a radius or a diameter?

tvillegas98:
> > Now projectiles explodes in player's current position
> 
> this doesn't seem to look proper. I recommend using the client to determine location for missile explosion animation. Also it probably should not be centered to the char but rather appear in the intersection where the collision occurred

I agree but probably they will need an entire visual refactor that it is not being addressed here, I can leave it as it is now

ncontinanza (approved):
LGTM!

PR #813, opened 2023-08-24
Feedback for basic skill

### LINKED ISSUE #810: Feedback for basic skill
When the Basic Attack button is pressed and held, there should be a feedback on the/those players that will be affected by the attack, if there's any in range. The attack is performed once the button is released. 

We'll begin with Muflus, and then extend to the rest of the characters.

ncontinanza:
Closes #810 

This PR implements a new feedback that is visualized on all those enemies that are in the range of a basic attack or Muflus' barrel roll or H4ck's Multishot

Steps to test:
1. Start a new game with at least two players (make sure to have at least one Muflus and one H4ck).
2. Use Muflus to get near to one of the other characters. Tap and hold the basic attack button, and notice how your enemy changes color when it's in your attack range.
3. Repeat for the barrel roll skill and the multishot skill.
4. Repeat for the basic attack of any other character.

Notes:
1. You currently won't see this feedback for D'Agna, as its basic attack is not yet configured for this purpose. But, if you want to verify that it will work when this is done, you can set a value for the `Skill Circle Radius` attribute in the `DAgna_Scherzo` scriptable object, and also set an Angle for this skill in the `skills.json` file.
2. For H4ck's case, the feedback will be shown only when the enemy is reachable in the SkillRange Area (that's the purple circle that is rendered at H4ck's feet when you tap and hold the basic attack button).
3. The current feedback we're using is temporary, and we'll make it nicer and more noticeable in a future PR.


client/Assets/Scripts/Utils/Utils.cs
```diff
@@ -66,4 +66,18 @@ string resource
...
+            Renderer renderer = character.CharacterModel.transform
+                .GetChild(i)
+                .GetComponent<Renderer>();
+            if (renderer)
+            {
+                renderer.material.color = color;
+            }
+        }
+    }
 }
```
BertovDev, line 83: I know this is a first iteration but I believe this would be better to have it in `PlayerFeedbacks`.

What do you think?

client/Assets/Scripts/Battle.cs
```diff
@@ -835,32 +835,10 @@ float characterSpeed
...
-                if (renderer)
-                {
-                    renderer.material.color = Color.white;
-                }
-            }
-        }
+        Color color = playerUpdate.Effects.ContainsKey((ulong)PlayerEffect.OutOfArea)
+            ? Color.magenta
+            : Color.white;
+        Utils.ChangeCharacterMaterialColor(character, color);
```
BertovDev, line 841: Maybe better to have a method in `PlayerFeedbacks`?

What do you think?

client/Assets/Scripts/Utils/Utils.cs
ncontinanza: I think it would lead to some coupling problems, as this is not only being used in the Battle class (where it would be easy to implement it using its instance of PlayerFeedbacks), but also in the CustomInputManager.

client/Assets/Scripts/Battle.cs
ncontinanza: This one is definitely better.

ncontinanza: Same as before

client/Assets/Scripts/AimDirection.cs
```diff
@@ -115,6 +112,33 @@ public void SetConeIndicator()
...
+
+        // Calculate the sides of the triangle
+        Vector3 sideAB = vertexB - vertexA;
+        Vector3 sideAC = vertexC - vertexA;
+
+        // return the bisector direction of the triangle's vertex angle
+        return (sideAB.normalized + sideAC.normalized).normalized;
+    }
+
+    public bool isInsideCone(GameObject player)
```
mgrunwaldt, line 133:
```suggestion
    public bool IsInsideCone(GameObject player)
```
functions in c# should start with capital letter, after the change is done please check where it's used

mgrunwaldt (changes requested):

client/Assets/Scripts/CustomInputManager.cs
```diff
@@ -477,4 +505,66 @@ public void ToggleCanceled(bool value)
...
+        return 0f;
+    }
+
+    private bool PlayerIsInSkillRange(GameObject p, Skill skill)
+    {
+        if (skill.GetSkillName() == "MULTISHOT")
+        {
+            return PlayerIsInSkillDirectionRange(p, skill);
+        }
+        return PlayerIsInSkillProximityRange(p, skill);
```
BertovDev, line 542: Same here, to avoid multiple returns

```
  bool isMultiShot = skill.GetSkillName() == "MULTISHOT";

	return isMultiShot
            ? PlayerIsInSkillDirectionRange(p, skill)
            : PlayerIsInSkillProximityRange(p, skill);
```

BertovDev (changes requested):
Awesome!! some minor changes. Everythings else is amazing!

BertovDev (dismissed):
Stunning! 🤙

mgrunwaldt:
Great work! We should also add feedback for H4CK Paralyze Ability
image

mgrunwaldt (dismissed):

client/Assets/Scripts/CustomInputManager.cs
```diff
@@ -296,7 +296,11 @@ public void ExecuteTapSkill(Skill skill)
         HideSkillRange();
     }
 
-    private void MapDirectionInputEvents(CustomMMTouchJoystick joystick, Skill skill)
+    private void MapDirectionInputEvents(
+        CustomMMTouchButton button,
+        CustomMMTouchJoystick joystick,
```
AngieDutra, line 301: All joysticks are buttons, so I believe you could only use buttons and then use GetComponent<CustomMMTouchJoystick>() when needed

client/Assets/Scripts/CustomInputManager.cs
```diff
@@ -483,4 +511,59 @@ public void ToggleCanceled(bool value)
...
+            if (PlayerIsInSkillRange(p, skill))
+            {
+                inRangeTargets.Add(p);
+            }
+        });
+        return inRangeTargets;
+    }
+
+
+    private bool PlayerIsInSkillRange(GameObject p, Skill skill)
```
AngieDutra, line 530:
```suggestion
    private bool PlayerIsInSkillRange(GameObject player, Skill skill)

```

AngieDutra: I would to this with all the following `p`

AngieDutra (changes requested):

client/Assets/Scripts/CustomInputManager.cs
```diff
@@ -175,10 +175,9 @@ public IEnumerator ShowInputs()
 
     public void AssignSkillToInput(UIControls trigger, UIType triggerType, Skill skill)
     {
-        CustomMMTouchJoystick joystick = mobileButtons[
-            trigger
-        ].GetComponent<CustomMMTouchJoystick>();
-        CustomMMTouchButton button = mobileButtons[trigger].GetComponent<CustomMMTouchButton>();
+        CustomMMTouchButton button = mobileButtons[trigger];
+        CustomMMTouchJoystick joystick = button.GetComponent<CustomMMTouchJoystick>();
+        //CustomMMTouchButton button = mobileButtons[trigger].GetComponent<CustomMMTouchButton>();
```
AngieDutra, line 180: I think you forgot to delete this comment. Other than that, it looks great!

AngieDutra (changes requested):

client/Assets/Scripts/Battle.cs
```diff
@@ -834,32 +834,10 @@ float characterSpeed
...
```
BertovDev, line 840: This line is droping FPS by 15-30. This is happening becuase it is been called every frame for each player no matters if the color had changed or not. So its consuming a lot of resources. 
( Also its doing a For loop inside a Update and passing through each transform child , really expensive method )

A solution to this could be only calling this method for the players that have their material changed.

BertovDev (changes requested):

client/Assets/Scripts/Battle.cs
ncontinanza: Nice catch! I made a similar optimization for the poisoned feedback as well.

BertovDev (commented):
Nice! one last thing. I found that the method `ShowSkillRange` in `CustomInputManager` its also killing the performance a little bit. This can be easily fixed but I know that method was not implemented in the PR. 
So what do you think, fix it in this PR o make a new PR with the fix?
I can commit the fix in any of the options.

---DISMISSED---

BertovDev (approved):

AngieDutra (approved):
LGTM

PR #814, opened 2023-08-24
Play sfx according to the player distance

### LINKED ISSUE #812: [REFACTOR] Play sfx according to the player distance
## 🚧 Refactor

We should only play sfx that are triggered near to the players position.
There will be some exceptions in the future but at the moment every skill sfx should follow this.

### Observations
- Let's start using the field of view as an approximation of how far we should play a sfx.
- 3D sounds: players should be able to identify from which direction a sfx was played.

BertovDev:
This PR adds abilities sfx sounds by proximity

- New `3DSoundManager` prefab.You can find it in each character prefab. It contains a MMF_PLAYER component to control the SoundManager sounds of the abilities. Also contains a `Sound3DManager` with 2 methods:

`SetSfxSound(AudioClip sfx)` Sets the clip to play in the MMF_PLAYER
`PlaySfxSound()` Plays the MMF_PLAYER feedbacks.

and finally in SKill.cs instead of using` PlayAbilityStartSfx` from CharacterAbility we use `PlaySfxSound` the method named before from the `Sound3DManager`

closes #812


ncontinanza (dismissed):
Loud and clear 👌

mgrunwaldt (changes requested):

ncontinanza (approved):

mgrunwaldt (approved):

Issue #777, opened 2023-08-10
[REFACTOR] Adjust input buttons position


saugon:
## 🚧 Refactor

The current input layout makes somewhat difficult to see the battlefield, considering that fingers are constantly placed on top of them.

Possible solutions:
1. Lower the buttons.
2. Reduce button size (specially skill basic).

### References
![IMG_0018](https://github.com/lambdaclass/curse_of_myrra/assets/6811225/dc5ff0ac-da54-403c-8aaf-0f6c0eb36180)
![IMG_0019](https://github.com/lambdaclass/curse_of_myrra/assets/6811225/e56494d1-d518-4ea6-9642-d65f9d0a3428)
![IMG_0020](https://github.com/lambdaclass/curse_of_myrra/assets/6811225/087822a7-45c3-464e-9efd-122ed9bc72a7)
![comparison](https://github.com/lambdaclass/curse_of_myrra/assets/6811225/265e1c16-e979-4d97-bc14-ba9f28e53d54)

PR #817, opened 2023-08-25
Change input size and position

AngieDutra:
This PR changes inputs position and size, based on issue #777 . Build it and test it on phone


BertovDev (changes requested):
Good! I believe the diantance of the cancel button is too much from the other buttons.It is intensional or nor?

BertovDev (approved):
Better for playability, awesome😎🤙

saugon (approved):
Nice!

PR #825, opened 2023-08-25
Warning about mismatching client and server hashes

### LINKED ISSUE #674: [FEATURE] Warning before joining lobbies with a different srv<>clt hash
Motivation:
- To make sure the game works
- To avoid false bug reports that are caused by the version mismatch

ncontinanza:
Closes #674 

When the server and client versions do not match, a warning is shown after trying to join a lobby, create a new game or a quickgame.

### DIFF
--- a/client/Assets/Scripts/LobbyConnection.cs
+++ b/client/Assets/Scripts/LobbyConnection.cs
@@ -45,6 +45,8 @@ public class LobbyConnection : MonoBehaviour
     public string ongoingGameDescription = "Do you want to reconnect to the game?";
     public string connectionTitle = "Error";
     public string connectionDescription = "Your connection to the server has been lost.";
+    public string versionHashesTitle = "Warning";
+    public string versionHashesDescription = "Client and Server version hashes do not match.";
 
     WebSocket ws;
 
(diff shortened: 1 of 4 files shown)


client/Assets/Scripts/LobbyConnection.cs
```diff
@@ -167,15 +169,25 @@ private void MaybeReconnect()
...
     public void ConnectToLobby(string matchmaking_id)
     {
+        ValidateVersionHashes();
         ConnectToSession(matchmaking_id);
         LobbySession = matchmaking_id;
     }
 
+    private void ValidateVersionHashes()
+    {
+        if (serverHash.Trim() != GitInfo.GetGitHash().Trim())
```
BertovDev, line 185: There is a problem here. When you change the server to Brasil or Europe the `serverHash` string is null.

client/Assets/Scripts/LobbyConnection.cs
```diff
@@ -270,6 +283,7 @@ IEnumerator GetLobbies()
                         webRequest.downloadHandler.text
                     );
                     lobbiesList = response.lobbies;
+                    serverHash = response.server_version;
```
BertovDev, line 286: `response.server_version ` is returning null in Europe and Brasil servers

BertovDev (changes requested):

client/Assets/Scripts/LobbyConnection.cs
BertovDev: Nevermind these servers do not have the changes deployed. I thinkg anything else its right

client/Assets/Scripts/LobbyConnection.cs
```diff
@@ -45,6 +45,8 @@ public class LobbyConnection : MonoBehaviour
     public string ongoingGameDescription = "Do you want to reconnect to the game?";
     public string connectionTitle = "Error";
     public string connectionDescription = "Your connection to the server has been lost.";
+    public string versionHashesTitle = "Warning";
+    public string versionHashesDescription = "Client and Server version hashes do not match.";
```
BertovDev, line 49: Do we really need these variables to be `public`? 
I believe they could all be `private`. 

Also in the future it will be better to have them in a different file, something like : ClientMessages or something like that

BertovDev (changes requested):

client/Assets/Scripts/LobbyConnection.cs
ncontinanza: phew!

client/Assets/Scripts/LobbyConnection.cs
ncontinanza: Actually they could be constants, I'll make this change.

BertovDev (approved):
Awesome!

AminArria (approved):

PR #865, opened 2023-09-06
Prevent the screen from turning off in case of inactivity

ncontinanza:
Ditto.

Make sure to try it out in the lobbies, lobby, character selection and battle screen.


AngieDutra (approved):
Tested on my phone and it works like a charm!

mgrunwaldt (approved):
Tested in iOS, works perfectly

Issue #744, opened 2023-08-05
[DESIGN] Reduce move-pad opacity


marco-paulucci:
## 🎨 Design

Reduce the opacity of the left pad to 0.3-0.5 and increase it slightly when active (+.1)

PR #878, opened 2023-09-08
Left joystick alpha changes

AngieDutra:
This PR solves #744 
To test it check the alpha change when using the left / movement joystick

it should start more transparent and it should have more opacity when using it.

### DIFF
--- a/client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
+++ b/client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
@@ -9,24 +9,38 @@ public class LeftMMTouchJoystick : MMTouchRepositionableJoystick
 {
     float positionX;
     float positionY;
+
     protected override void Start()
     {
         base.Start();
         _initialPosition = BackgroundCanvasGroup.transform.position;
     }
-    private Vector3 clampJoystickPositionToScreen(PointerEventData eventData)
+
+    private Vector3 ClampJoystickPositionToScreen(PointerEventData eventData)
     {
-        if (eventData.position.y < GetComponent<RectTransform>().position.y + BackgroundCanvasGroup.GetComponent<RectTransform>().sizeDelta.y / 2)
+        if (
+            eventData.position.y
+            < GetComponent<RectTransform>().position.y
+                + BackgroundCanvasGroup.GetComponent<RectTransform>().sizeDelta.y / 2
+        )
         {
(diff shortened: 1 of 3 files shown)


ncontinanza (approved):
Nice!

client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
```diff
@@ -35,21 +49,33 @@ private Vector3 clampJoystickPositionToScreen(PointerEventData eventData)
         _newPosition = new Vector3(positionX, positionY, 0f);
         return _newPosition;
     }
+
+    public void SetOpacity()
+    {
+        BackgroundCanvasGroup.alpha = 0.3f;
```
mgrunwaldt, line 55: Lets avoid using magic variables. Instead of using 0.3f here I'd define a const and then use it here

client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
```diff
@@ -35,21 +49,33 @@ private Vector3 clampJoystickPositionToScreen(PointerEventData eventData)
...
+    }
+
+    public void UnsetOpacity()
+    {
+        BackgroundCanvasGroup.alpha = 0.4f;
```
mgrunwaldt, line 60: same comments as above

mgrunwaldt (changes requested):

client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
BertovDev, line 1: Is this a custom Script or it is part of the engine?

AngieDutra: It's custom but it reverences `MMTouchRepositionableJoystick`

mgrunwaldt (approved):

client/Assets/Scripts/UI/LeftMMTouchJoystick.cs
```diff
@@ -9,24 +9,40 @@ public class LeftMMTouchJoystick : MMTouchRepositionableJoystick
 {
     float positionX;
     float positionY;
+    float initialJoystickOpacity = 0.3f;
+    float pressedJoystickOpacity = 0.4f;
```
mgrunwaldt, line 13: these variables won't change in runtime, I'd define them as a const

BertovDev (approved):
Awesome

- [x] Tested in Android

ncontinanza (approved):
LGTM! 🚀

PR #882, opened 2023-09-11
Cel Shading + Lighting improvements

saugon:
### What's new?
1. Added cel shading materials for each character.
2. Added post processing effects.
3. Tweaked lighting.

### What's newer?
1. Character feedback refactor: merged overlay and player feedback scripts into one.
2. Improved overlay visual feedback.
3. Added "in the zone" feedback.

#### Cel shading + lighting: before and after
Screenshot 2023-09-11 at 19 44 13
Screenshot 2023-09-11 at 19 41 06
#### Bloom effect
Screenshot 2023-09-11 at 19 42 12
#### Zone effect
Screenshot 2023-09-19 at 18 14 25


ncontinanza:
I noticed the following behaviour when playing with H4ck in my mobile phone and Muflus in my Unity editor. Am I missing some step before testing, or is there a bug?

https://github.com/lambdaclass/curse_of_myrra/assets/17294394/3523282d-c29e-464f-986c-c298dd065a0f


https://github.com/lambdaclass/curse_of_myrra/assets/17294394/23e5205c-bd36-4ae6-a4d2-c64a060d7694

BertovDev (changes requested):
inside `CharacterBase` in the `PlayerName` prefab there is a `UmaMarks` gameObject with a missing script and a empty image component. I reload unity and remain still. It is expected to add a new functionality later?
Captura de pantalla 2023-09-20 a la(s) 10 35 35

AngieDutra (changes requested):
https://github.com/lambdaclass/curse_of_myrra/assets/82987608/06e4c141-ce9f-4db5-8260-530d718ac66c

https://github.com/lambdaclass/curse_of_myrra/assets/82987608/faf3403b-37b8-47db-b9c0-51fce46ad7b1

- Players look too bright and in hack's case it dificults to see the damage feedback
- The red feedback in the screen when a player enters the kill zone starts a bit too late. Maybe we could make the first one appear a bit faster and if the damage is too great then the second one appears. I would check if it's possible to add some transition to it too.
- In main players used to turn red if they were within the attack range from another player. Now they don't give any feedback in that case.
- Defeated splash shows the player without any color (only white)

mgrunwaldt (approved):

Issue #779, opened 2023-08-11
[BUG] Buttons in the lobbies scene remain clicked


ncontinanza:
## 🐛 Bug Report

When a pop up is activated (for example, when the connection to the server is lost), the buttons in the lobbies scene don't go back to their "not-pressed" state after being pressed.

![image](https://github.com/lambdaclass/curse_of_myrra/assets/17294394/de3f50a8-ec77-422d-be79-b4c03d011e2c)

A possible reason of this problem can be found [here](https://github.com/lambdaclass/curse_of_myrra/pull/726#issuecomment-1675062695)

Issue #664, opened 2023-07-27
[Bug] Constant refresh if user holds down the refresh button in lobby


AngieDutra:
This Issue extends to all buttons. When one button is pressed and the user holds the finger in that position, if in the next scene there is a button in the same position the finger is pressing, the new button will trigger although that wasn't the users intention.

Instead of re rendering the scene we should only re render the components needed and and make it error proof with a check on release

PR #884, opened 2023-09-12
Refresh refactor

AngieDutra:
This PR solves issues #664 and #779 
It optimizes the way we refresh our lobbies list and changes the buttons styles (in lobbies scene) to fix ui and ux problems.

Issues to check if they were solved:
- Hold your finger in the Refresh button. It should no longer refresh constantly. Instead, it should refresh when you stop pressing on it.
- If you press the Refresh button it should no longer toggle as much between the empty list indicator and the lobbies. Instead it should show either one or the other until your next refresh. 
_Two things to take into account: now it may not show you a lobby if it took too long to fetch it. That is expected and it works like that in other games too. In that case you should refresh again; If you spam the Refresh button it may occur that a lobby list is shown after an empty list indicator. This is also expected behavior._
- If an error is shown after or while a button is pressed, your buttons should no longer stay in a pressed state.
_You can trigger a pop up to test this by starting a game with a different hash to trigger a warning or pressing the server button and display the server options_

_Keep in mind the button changes were only made in the lobbies scene since a refactor of buttons styles is coming up_

### DIFF
--- a/client/Assets/Scripts/UIManager.cs
+++ b/client/Assets/Scripts/UIManager.cs
@@ -14,7 +14,6 @@ public class UIManager : MonoBehaviour
 
     [SerializeField]
     GameObject noLobbiesText;
-    List<string> lobbiesList;
 
     void Start()
     {
@@ -47,8 +46,7 @@ public class UIManager : MonoBehaviour
 
     void GenerateList()
     {
-        lobbiesList = new List<string>();
-        lobbiesList = LobbyConnection.Instance.lobbiesList;
+        List<string> lobbiesList = LobbyConnection.Instance.lobbiesList;
         lobbiesList.Reverse();
         lobbiesList.ForEach(el =>
         {
--- a/client/Assets/Scripts/UIManager.cs
+++ b/client/Assets/Scripts/UIManager.cs
@@ -16,32 +16,48 @@ public class UIManager : MonoBehaviour
     GameObject noLobbiesText;
 
     bool lobbiesEmpty = true;
-    bool gamesEmpty = true;
+    public List<string> lobbiesList;
+    private IEnumerator init;
 
     void Start()
     {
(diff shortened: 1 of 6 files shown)


client/Assets/Scripts/UIManager.cs
```diff
@@ -14,34 +14,52 @@ public class UIManager : MonoBehaviour
...
+        });
     }
 
-    public void GenerateList(List<string> itemList, Object itemPrefab, Transform container)
+    void GenerateList()
     {
-        itemList.Reverse();
-        itemList.ForEach(el =>
+        lobbiesList = new List<string>();
+        lobbiesList = LobbyConnection.Instance.lobbiesList;
```
ncontinanza, line 51:
```suggestion
        lobbiesList = LobbyConnection.Instance.lobbiesList;
```

ncontinanza: Could we directly assign `LobbyConnection.Instance.lobbiesList` instead of creating a new List and overwriting it with the other one?

ncontinanza (changes requested):

BertovDev (changes requested):
Working good!
I believe you have 2 instances of `UIManager` in `LobbiesCameraUI.prefab` by mistake. Or it was intentionally?

- One instance in `OptionsContainer` and the other in `LobbiesContainer` prefabs

AngieDutra:
> Working good! I believe you have 2 instances of `UIManager` in `LobbiesCameraUI.prefab` by mistake. Or it was intentionally?
> 
> * One instance in `OptionsContainer` and the other in `LobbiesContainer` prefabs

It was a mistake! Good catch

BertovDev (approved):

ncontinanza (approved):

PR #901, opened 2023-09-15
Map all characters sfx and separate sound channels

### LINKED ISSUE #895: [FEATURE] Map sounds with skills
## 🚀 Feature

 Map sounds with each skill.

BertovDev:
This PR adds all the sounds for the current characters we have and each ability.

Also refactor the mute/unmute and the slider system of the settings. Separating channels in Music/Sfx.

To test:

- The state of the track (mute/unmute) should persist thought scenes.
- In the Battle Scene you should control both SFX and Music separately.
- Mute the Music with the mute/unmute image then while muted change the volume slider, finally unmute the volumen should be set in the position where you left the slider when it was muted.

closes #895 and #903

- [x] Tested in Unity Desktop
- [x] Tested in Android


client/Assets/Scripts/UI/ToggleAudio.cs
```diff
@@ -21,29 +21,34 @@ public class ToggleAudio : MonoBehaviour
...
+        muteButtonImage.overrideSprite = IsMuted(channel) ? mutedSprite : unmutedSprite;
     }
 
     void Update()
     {
-        if (!IsMuted(MMSoundManager.MMSoundManagerTracks.Master))
+        if (
+            volumeSlider
+            && (IsMuted(channel) && unmutedVolume != volumeSlider.value)
+            && volumeSlider.value > 0.0001f
```
mgrunwaldt, line 42: why not > 0?

mgrunwaldt (commented):
I have a few questions / comments, not on the code but on the result:

1. Muflus barrell roll should start before (not sure if this is from implementation or if the actual sound track takes too long to start)
2. Muflus leap should coordinate better, the sound is slow and then has a stomp, the stomp should sound when muflus touches the ground. Question: Does every leap take exactly the same time, regardless of the length of the leap?
3. The radius for when to listen to an opponent should be much shorter, I am listening to sounds of characters I don't see on the screen, this can become really confusing
4.

martinacantaro:
Comments on how we could improve the sfx after we reviewed this with @BertovDev 
https://www.notion.so/lambdaclass/Music-feedback-17fb02c093d040bdbe9a5878e02d7a31

client/Assets/Scripts/UI/ToggleAudio.cs
ncontinanza: @mgrunwaldt @BertovDev I recall this is a value defined in some TopDown Engine classes to represent the minimum volume that a track can reach (for some reason, it's never 0). However, it may lead to some confusion. I think it will be better if we define our own constant here, and use that value wherever we need to mute a track.

ncontinanza (changes requested):

ncontinanza (approved):

ncontinanza (approved):

agustinesco (changes requested):
When you enter to a game when you've muted the game through the icon in the menus the ingame volumes seems a bit weird

image

the music slider is correctly muted but the sfx slider is at max and the icon seems muted

BertovDev:
> When you enter to a game when you've muted the game through the icon in the menus the ingame volumes seems a bit weird
> Screenshot 2023-10-20 at 11 48 04 image
> 
> the music slider is correctly muted but the sfx slider is at max (you can hear the sfx) and the icon seems muted

Currently we only have the option to mute the music in the lobbies and we could implement a sfx mute button also. Nice catch this could be a new issue.

agustinesco:
> > When you enter to a game when you've muted the game through the icon in the menus the ingame volumes seems a bit weird
> > Screenshot 2023-10-20 at 11 48 04 image
> > the music slider is correctly muted but the sfx slider is at max (you can hear the sfx) and the icon seems muted
> 
> Currently we only have the option to mute the music in the lobbies and we could implement a sfx mute button also. Nice catch this could be a new issue.

Oh okey , it was a misunderstanding. I though this button would mute every sound not only the music, thanks for explaining!

client/Assets/Scripts/UI/ToggleAudio.cs
```diff
@@ -58,28 +66,50 @@ public void Toggle()
...
         {
-            soundManager.SetVolumeMaster(newVolume);
+            case MMSoundManager.MMSoundManagerTracks.Music:
+                soundManager.SetVolumeMusic(newVolume);
+                break;
+            case MMSoundManager.MMSoundManagerTracks.Sfx:
+                soundManager.SetVolumeSfx(newVolume);
+                break;
         }
     }
```
agustinesco, line 107: I think we should update the volume sprite here too, because if we are muted and then increase only the volume the sprite doesn't change from the muted one

BertovDev: Awesome, another good catch.

agustinesco (approved):
LGTM!

client/Assets/Scripts/UI/ToggleAudio.cs
agustinesco: Nice!

ncontinanza (approved):

PR #907, opened 2023-09-18
Setup player abilities dynamically

BertovDev:
Since ScripteableObject lives in the editor, we couldnt change them values ar run-time. So the solution is to Instanciate a copy of them and use it.
With this PR we can instanciate a copy of the skills ScripteableObject maping the backend data from the Skills.json. 

It also destroy the instances when the player is dead. 

SO stands for ScripteableObject

How it works?
- We have the original Skills SO stored in a list inside the Character SO
- In the `CustomLevelManager` we iterate over them and create a clone of each one.
- Then we map the Skill.json data to the copies using the constructor `skillClone.InitWithBackend()`
- Then we assing them to the Skills
- Finally when the player is dead we `Destroy` each clone instance.

To test:
- Play a Game for example with H4ck
- Use the multishot skill and see the angle of the cone
- Exit the game ( but dont stop the playmode)
- Edit the Skill.json for the Multiushot skill. Increasing the angle.
- Start a new game again and check if the angle has changed.


client/Assets/Scripts/CoMCharacter.cs
```diff
@@ -11,11 +11,8 @@ public class CoMCharacter : ScriptableObject
     public Sprite blockArtwork;
     public Sprite characterPlayer;
     public GameObject prefab;
+    public List<SkillInfo> skills;
```
AngieDutra, line 14: You already had a list of this in line 24 of this file

AngieDutra (changes requested):
Also, I see this error and I would like to know if you see it too

https://github.com/lambdaclass/curse_of_myrra/assets/82987608/aae1ed9a-3b90-4a81-8bed-2c7877ab31de

AngieDutra (approved):
Great job!

ncontinanza (dismissed):
Awesome!

mgrunwaldt:
While testing this branch (after merging main to it) I don't see h4ck's projectiles:

https://github.com/lambdaclass/curse_of_myrra/assets/22064054/154ee10c-14ad-42b4-bc44-758405de34a6

This is the error it logs:
image
image

ncontinanza (approved):

ncontinanza (approved):

AngieDutra (approved):
Great work! LGTM ✨

PR #909, opened 2023-09-19
In-game UI refactor

AngieDutra:
## Changes
- In game UI
- Character selection UI
- Delete sprites and optimize the way we handle them
## How to test it
- In character selection go through each player and their skills to check they are working as expected (selected and not selected styles)
-   Use all skill inputs and check their cooldown and disable status is working


ncontinanza (dismissed):
LGTM! 🚀

ncontinanza (approved):

client/Assets/Scripts/UI/CustomMMTouchButton.cs
```diff
@@ -3,37 +3,55 @@
...
 
     public override void OnPointerUp(PointerEventData data)
     {
         newPointerTapUp.Invoke(skill);
         base.OnPointerUp(data);
+        foreach (Image image in images)
+        {
+            if (image.GetComponent<CustomMMTouchButton>() == null)
+            {
+                image.color = new Color(255f, 255f, 255f);
```
mgrunwaldt, line 40: 2 comments on this: 

1. you are creating a new color for each image, when the color is always the same. Couldn't you just initialize the color once and then use it without creating one each time?
2. Even though we didn't create a class for colors (that will be the ideal case) lets try to avoid defining colors (or any magic numbers) in the code, this is easy because its white, but if it's a different combination of rgb it would be hard for someone reading the code to know what color you're creating

To solve both problems at the same time you can create a variable for this same color at the beginning of the class and then use it here

client/Assets/Scripts/UI/CustomMMTouchButton.cs
```diff
@@ -3,37 +3,55 @@
...
 
     public override void OnPointerPressed()
     {
         newPointerTapDown.Invoke(skill);
         base.OnPointerPressed();
+        foreach (Image image in images)
+        {
+            if (image.GetComponent<CustomMMTouchButton>() == null)
+            {
+                image.color = new Color(0.5f, 0.5f, 0.5f);
```
mgrunwaldt, line 53: same comment as above

mgrunwaldt (approved):

PR #912, opened 2023-09-20
Uma Yugen's Mark

### LINKED ISSUE #803: [FEATURE] Uma - Yugen's Mark
## 🚀 Feature

Active 1, aoe poison attack

- [ ] Throw a poison breath in a semi-short (=SkillRange) ranged cone
- [ ] Afftected targets are given Mark2 for (Duration=6000ms) time
- [ ] Deals poison damage. "Damage" param should represent the total damage dealt over the period in ms. This might have considerable presicion loss, see #633
- [ ] Poison damage duration is the same as the mark, but doesn't need to be coupled
- [ ] Poison cannot kill! it reduces HP to 1 at most. 
- [ ]  Mark2 client feedback
- [ ] 4500ms cd

⚠️ Marks Info
- Applying the mark on a target which already has that mark resets the duration, rather than stack
- Only Uma and targets affected by Marks should see their feedback.
- Uma should not see nor interact with Marks from other Umas in any way

ncontinanza:
Closes #803 
This PR implements several changes on Uma's Yugen's Mark (skill 1).

Things to test:
- Yugen's Mark has a range and an angle, so there is a cone that aids the player as there we did with H4ck's Multishot skill. Don't mind the in-range feedback, as it will be removed as a part of another PR.
- All those players inside the cone will be affected by this skill.
- The duration of this mark is of 6000ms.
- The cooldown lasts for 4500ms.
- The feedback for this Mark is a drop above the marked character's name, and its healthbar turning green. If more than one kind of mark is applied (for example, Yugen's + Elnar's), more than one drop will be shown.
- The damage that this effect applies to the receiver is not applied all at once. You should see it evenly distributed during the period of time that the Mark lasts.
- A player can't be killed by the Yugen's Mark. If the player's health is 1hp, the Yugen's Mark will stop causing damage.


server/native/gamestate/src/player.rs
AminArria, line 1: This changes need to go to [lambda_game_engine](https://github.com/lambdaclass/lambda_game_engine)

AminArria (changes requested):
Changes to the NIFs need to go to [lambda_game_engine](https://github.com/lambdaclass/lambda_game_engine)

server/native/gamestate/src/player.rs
ncontinanza: Done!

client/Assets/Scripts/Battle.cs
```diff
@@ -32,7 +32,7 @@ public class Battle : MonoBehaviour
 
     private Loot loot;
     private bool playerMaterialColorChanged;
-    private bool healthBarColorChanged;
+    private bool healthBarIsGreen;
```
mgrunwaldt, line 35: I think we dont use this anymore

client/Assets/Scripts/CharacterFeedbackManager.cs
```diff
@@ -7,35 +7,29 @@ public class CharacterFeedbackManager : MonoBehaviour
...
     public UmaMarks umaMarks;
 
-    public ulong currentMark = 0;
+    IDictionary<ulong, ISet<PlayerEffect>> playersMarks =
+        new Dictionary<ulong, ISet<PlayerEffect>>();
 
-    List<ulong> marksApplied = new List<ulong>();
-
-    public void DisplayUmaMarks(ulong markId)
+    public void DisplayEffectMarks(ulong playerId, PlayerEffect effect)
```
mgrunwaldt, line 13: Now that we are sending the mark we want to display, I would rename to DisplayEffectMark

client/Assets/Scripts/CharacterFeedbackManager.cs
```diff
@@ -7,35 +7,29 @@ public class CharacterFeedbackManager : MonoBehaviour
...
-            marksApplied.Add(markId);
+            playersMarks.Add(playerId, new HashSet<PlayerEffect>() { effect });
         }
-        UpdateMarkImage(marksApplied.Count);
-        currentMark = markId;
+        UpdateMarkImage(playersMarks[playerId].Count);
     }
 
-    public void RemoveMarks(ulong markId)
+    public void RemoveMarks(ulong playerId, PlayerEffect effect)
```
mgrunwaldt, line 27: same comment as above

mgrunwaldt (approved):

martinacantaro (approved):
Playtested and approved!

PR #921, opened 2023-09-28
Change default server from Localhost to Brazil

### LINKED ISSUE #916: [REFACTOR] Make default server Brazil
## 🚧 Refactor

Currently, the default server for when you open the app is localhost. This causes that the error message appears always when you play from a build, since localhost isn't running in the phone. This should be changed so that the default server is brazil

ncontinanza:
Closes #916 

Change default server from Localhost to Brazil, to avoid getting the "lost connection" pop up when opening the game from a mobile device.

Also, several variable and attributes names that were in snake_case were changed to camelCase to follow C# naming conventions

### DIFF
--- a/client/Assets/Scripts/InputText.cs
+++ b/client/Assets/Scripts/InputText.cs
@@ -18,6 +18,6 @@ public class InputText : MonoBehaviour
 
     public void Init()
     {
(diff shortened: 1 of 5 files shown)


mgrunwaldt (approved):

BertovDev (approved):
Nice

PR #924, opened 2023-10-02
Start screen

AngieDutra:


AngieDutra:
This will be solve with and updated version of main in [this PR](https://github.com/lambdaclass/curse_of_myrra/pull/952)

PR #951, opened 2023-10-06
Crate and Spawn sfx

BertovDev:
This PR adds sfx for the crates and spawn


ncontinanza (approved):

client/Assets/Scripts/Loot.cs
```diff
@@ -28,30 +34,79 @@ void Start()
...
+    }
+
+    private bool ExistInLoots(ulong id)
+    {
+        foreach (var loot in loots)
+        {
+            if (loot.id == id)
+                return true;
+        }
+        return false;
```
agustinesco, line 88: Can we do something like this here?

```
return loots.Any(l => l.id == id)
```

client/Assets/Scripts/Loot.cs
```diff
@@ -28,30 +34,79 @@ void Start()
...
+                    RemoveLoot(loot.id);
+                }
+            });
     }
 
     private void RemoveLoot(ulong id)
     {
-        GameObject lootObject = loots[id];
+        GameObject lootObject = GetLootObject(id);
+        string type = GetLootType(id);
```
agustinesco, line 69: Since we already have the lootObject can't we do `lootObject.type` here? if so maybe we could remove `GetLootType` method too

agustinesco (changes requested):

client/Assets/Scripts/Loot.cs
BertovDev: Yess, this is a much cleaner solution. Great one!!

client/Assets/Scripts/Loot.cs
BertovDev: No we cant, becuase lootObject is a GameObject and it dont have the field type. But instead we can do a method called `GetLoot` that returns a `LootItem` which contains both lootObject and type. 

Nice catch!!

client/Assets/Scripts/Loot.cs
agustinesco: Nice!

client/Assets/Scripts/Loot.cs
agustinesco: Sounds good

agustinesco (approved):
LGTM! sounds are working as expected

ncontinanza (approved):

PR #954, opened 2023-10-09
Main screen

AngieDutra:
The new main screen should look like this:
Refe-New
 Check:
- It is usable from title screen to lobbies screen
- The animations size is okey

NOTE: I did a refactor for ui animations that are based off 3d models so you should also check the death splash (victory and defeat animations for each character) are working as expected


BertovDev (changes requested):
This is awesome and I believe we should not be using Character `prefabs` to be displayed in the screen since their contains scripts we dont need. We should use the `Models`. 

What do you think?

ncontinanza (approved):
LGTM!

BertovDev (dismissed):

ncontinanza (changes requested):
Looking great! Just an issue in Uma's victory screen, the animation is frozen after the first few seconds

https://github.com/lambdaclass/curse_of_myrra/assets/17294394/6b687cdd-8368-4e25-9f32-784b45636618

ncontinanza (approved):
Working great!

BertovDev (approved):
Looks great!

- [x] Tested in Unity Desktop
- [x] Tested in Android

PR #973, opened 2023-10-12
EngineRunner h4ck skills

AminArria:
Related to https://github.com/lambdaclass/lambda_game_engine/pull/17

This update implemetents the following
- Shooting skills
- Moving and "applying" projectiles during game ticks
- Applying effects during game ticks

Pending
- [x] Update `mix.lock` once merged https://github.com/lambdaclass/lambda_game_engine/pull/17

## Testing this PR

In `config.exs` change the config to
```
config :dark_worlds_server, DarkWorldsServer.Engine.Runner,
  process_priority: :high,
  use_engine_runner: true  ## <- Setting true enables EngineRunner to replace Runner
```

You can only use H4ck for testing purposes and only 2 skills will work, the "basic attack" one which does a multi shot and skill number 2 (middle of the three buttons), that one will do another short range shot which is a poison dart that will put a poison effect (some damage to the player for 5 secs every second)

Bots can be manually spawned if you want, you will need the `EngineRunner` PID which can be seen in the log line

```
[info] Switching to engine_runner #PID<0.642.0>
```

And then you only need to invoke this command with that PID

```
iex(21)> DarkWorldsServer.Engine.EngineRunner.join(pid(0, 642, 0), "bot1", "h4ck")
```

`"bot1"` can be any string, but `"h4ck"` has to remain because no other character is configured in config at the moment


agustinesco (approved):
LGTM! projectiles seems to be working as expected

ncontinanza (changes requested):
I'm getting this error when spawning bot and attacking it with the new skills:

```
Assertion failed on expression: 'CompareApproximately(SqrMagnitude(result), 1.0F)'
UnityEngine.Quaternion:Internal_FromEulerRad (UnityEngine.Vector3)
AimDirection:Rotate (single,single,Skill) (at Assets/Scripts/AimDirection.cs:69)
CustomInputManager:ShowAimDirectionSkill (CustomMMTouchJoystick) (at Assets/Scripts/CustomInputManager.cs:318)
UnityEngine.Events.UnityEvent`1<CustomMMTouchJoystick>:Invoke (CustomMMTouchJoystick)
CustomMMTouchJoystick:OnPointerDown (UnityEngine.EventSystems.PointerEventData) (at Assets/Scripts/UI/CustomMMTouchJoystick.cs:18)
UnityEngine.EventSystems.EventSystem:Update ()
```

Also, I don't get the poisoned feedback when using the paralyze skill with the new settings, should that happen?

AminArria:
> I'm getting this error when spawning bot and attacking it with the new skills:

@ncontinanza no clue about that error, as in I don't understand why it is happening. Might be something about how the button works in the client

> Also, I don't get the poisoned feedback when using the paralyze skill with the new settings, should that happen?

Is there a poison feedback? Also, a lot of those things are likely to not work and this is expected, the goal here is to have things working in the engine and as much as possible in the client, but broken things are expected

ncontinanza:
Great, I will approve the PR right now then.

> Is there a poison feedback?
Yes, currently we're changing the health bar color to green while the Poisoned effect lasts.

ncontinanza (approved):

PR #983, opened 2023-10-18
Fix spawn VFX 

### LINKED ISSUE #982: [BUG] Spawn vfx activate when player appears
## 🐛 Bug Report

Should be once in the Init and then never again.

BertovDev:
Desactivate spawn VFX after the first time it is played.

closes #982

### DIFF
--- a/client/Assets/Scripts/CharacterBase.cs
+++ b/client/Assets/Scripts/CharacterBase.cs
@@ -25,8 +25,9 @@ public class CharacterBase : MonoBehaviour
 
     IEnumerator activateSpawnFeedback()
     {
+        float lifeTime = spawnFeedback.GetComponent<VisualEffect>().GetFloat("LifeTime");
         spawnFeedback.SetActive(true);
-        yield return new WaitForSeconds(2f);
+        yield return new WaitForSeconds(lifeTime);
         spawnFeedback.SetActive(false);
     }
 
--- a/client/Assets/Scripts/CharacterBase.cs
+++ b/client/Assets/Scripts/CharacterBase.cs
@@ -1,6 +1,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.VFX;
 
 public class CharacterBase : MonoBehaviour
 {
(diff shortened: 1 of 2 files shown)


client/Assets/Scripts/CharacterBase.cs
```diff
@@ -1,6 +1,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.VFX;
```
tkz00, line 4: This using directive is unnecessary.

tkz00 (changes requested):
While this mostly solves the issue at hand, the bug still arises when you find another player before the two seconds that it takes for the spawn feedback game object to deactivate, and in that case not only is the animation is still played when you find the other player, instead of at the start, but also it will be ended abruptly before time while it's still playing. Most importantly it can also generate other problems in the future, e.g. if we change the spawn animation to last more than 2 seconds, then the spawn feedback object will be deactivated before the animation fully completes, ending it before time. (In the attached video I changed the yield return to .5f seconds to show the problem I'm describing)


https://github.com/lambdaclass/curse_of_myrra/assets/67983386/c6473771-4868-4505-9ef4-c1159ef8b34f

tkz00 (approved):

BertovDev:
> While this mostly solves the issue at hand, the bug still arises when you find another player before the two seconds that it takes for the spawn feedback game object to deactivate, and in that case not only is the animation is still played when you find the other player, instead of at the start, but also it will be ended abruptly before time while it's still playing. Most importantly it can also generate other problems in the future, e.g. if we change the spawn animation to last more than 2 seconds, then the spawn feedback object will be deactivated before the animation fully completes, ending it before time. (In the attached video I changed the yield return to .5f seconds to show the problem I'm describing)
> Screen.Recording.2023-10-18.at.17.29.41.mov

You are totally right, I have changed the values to a dinamic one depending of the vfx lifetime. Although i can´t recreate what you said here. 

> the bug still arises when you find another player before the two seconds that it takes for the spawn feedback game object to deactivate

Good catch, any other suggestion?

ncontinanza (approved):
LGTM!

PR #986, opened 2023-10-19
Trim custom server address.

### LINKED ISSUE #940: [BUG] Trim text field for custom server address
## 🐛 Bug Report

It's quite common to copy paste a server's address leaving a blank space at the end of the string. This causes an error that could be easily avoided if we trimmed the string before attempting to connect.

tkz00:
Trims the start and end of the select server custom input in the lobby selection screen.

Closes #940


BertovDev (approved):
🤙

ncontinanza (approved):

PR #1003, opened 2023-10-19
Reconnect improvements

### LINKED ISSUE #786: [FEATURE] Re-connect improvements
## 🚀 Feature

- [x] there are some scenarios in which you are left inside the game, but unable to perform any actions. It would be nice to automatically perform a reconnect/refresh of the game (without prompt) if you are disconnected in the "playing" scene
- [ ] you should not be prompted to re-join a game that is (a) already finished,
- [ ] (b) in which you are dead
- [ ] or (c) you explicitly exited

ncontinanza:
Closes #786 

With this PR, you should not see the reconnect pop up after manually exiting a match (whether you do it pausing the game and tapping on the exit game button, or doing it after you were killed and the match goes on).
You should still be able to reconnect if you lose connection, or close the app and open it again.


agustinesco (changes requested):
Tested the pr and the re-join option isn't being prompt when i manually exit the game as requested in the issue! but this makes me wonder, when do we want this option to appear? i exit'ed an on-goin game by killing the game pressing the "Play" button in the unity hud and when i go back to the lobbies the option isn't there

server/lib/dark_worlds_server_web/controllers/game_controller.ex
```diff
@@ -25,11 +26,21 @@ defmodule DarkWorldsServerWeb.GameController do
...
+          Enum.map(selected_characters, fn {id, name} -> %{character_name: name, id: id} end)
 
-        selections = Enum.map(selected_characters, fn {id, name} -> %{character_name: name, id: id} end)
         game_config = Enum.reduce(game_config, %{}, &transform_config/2)
-        server_hash = Application.get_env(:dark_worlds_server, :information) |> Keyword.get(:version_hash)
+
+        server_hash =
+          Application.get_env(:dark_worlds_server, :information) |> Keyword.get(:version_hash)
+
+        player_exited_game =
```
AminArria, line 40: Look at the `PlayerTracker` module, essentially what you should do is call something to trigger the removal of the player from

server/lib/dark_worlds_server/engine/runner.ex
```diff
@@ -234,6 +234,17 @@ defmodule DarkWorldsServer.Engine.Runner do
      }}
   end
 
+  def handle_cast(
+        {:play, _, %ActionOk{action: :exit_match, value: player_id}},
+        gen_server_state
+      ) do
+    {:ok, game} = Game.exit_game(gen_server_state.server_game_state.game, player_id)
```
AminArria, line 241: As I mentioned in the other PR, the engine should not have any notion about this kind of things

server/lib/dark_worlds_server_web/controllers/game_controller.ex
AminArria, line 1: I see a lot of formatting changes in this file, please check that whoever you format follows our config https://github.com/lambdaclass/curse_of_myrra/blob/main/server/.formatter.exs

AminArria (changes requested):
By the way, are you formatti

server/lib/dark_worlds_server_web/controllers/game_controller.ex
ncontinanza: Thanks, I hadn't thought of that. I used the remove_player function and afterwards check if the player exists to decide wether it should try to reconnect.

server/lib/dark_worlds_server_web/controllers/game_controller.ex
ncontinanza: Fixed!

AngieDutra (commented):
Some buttons like mute button and back button are not shoeing properly. Maybe you should revert the overrides and push those changes in Lobbies.unity

ncontinanza:
> Some buttons like mute button and back button are not shoeing properly. Maybe you should revert the overrides and push those changes in Lobbies.unity

Done! I tested it and everything seemed to be working fine in the Lobbies scene, but please double check

PR #1007, opened 2023-10-19
[GH-987] Make bots flee from harmful zone

### LINKED ISSUE #987: Bots must escape from the damage zone

agustinesco:
Closes #987 

The bots should run from the zone and do their best effort to stay in the safe zone now


server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -6,6 +6,9 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
   alias DarkWorldsServer.Engine.Runner
   alias LambdaGameEngine.MyrraEngine.RelativePosition
 
+  # This variable will dice how much time passes between bot decisions in milis
+  @decide_delay 500
```
AminArria, line 10: nit: add suffix `_ms` to the variable name

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -50,19 +52,32 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
     bot_state = get_in(state, [:bots, bot_id])
-    new_bot_state = decide_action(bot_id, state.players, bot_state)
-    state = put_in(state, [:bots, bot_id], new_bot_state)
-    {:noreply, state}
-  end
 
-  def handle_info({:do_action, bot_id}, %{bots_enabled: false} = state) do
-    Process.send_after(self(), {:do_action, bot_id}, state.game_tick_rate)
+    new_bot_state =
+      if bot_state && bot_state.alive do
```
AminArria, line 57: Checking `bot_state.alive` here is redundant because we already check it in `decide_action/4` 

This also removed the need to check `bot_state`, this will never be `nil` unless we made a bug

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -169,4 +196,16 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
     y = Nx.sin(angle) |> Nx.multiply(radius) |> Nx.to_number()
     {x, -y}
   end
+
+  def decide_state(_player_state, %{bots_enabled: false}) do
+    :nothing
+  end
+
+  def decide_state(player_state, _bots_genserver_state) do
```
AminArria, line 204: This should be called `decide_objective`, the word "state" is really overloaded and use too much so better to avoid it when we have clearer words

AminArria (changes requested):

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -50,19 +52,32 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
-    {:noreply, state}
-  end
 
-  def handle_info({:do_action, bot_id}, %{bots_enabled: false} = state) do
-    Process.send_after(self(), {:do_action, bot_id}, state.game_tick_rate)
+    new_bot_state =
+      if bot_state && bot_state.alive do
+        Process.send_after(self(), {:decide_action, bot_id}, @decide_delay)
+
+        game_state = Runner.get_game(state.game_pid) |> Map.get(:myrra_state)
```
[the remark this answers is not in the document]
agustinesco: sounds good, i will just keep the game in the genserver state and update it here.

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: You're right! it's starting to get confusing to call everything "state". Changed

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: Sounds good

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: This check is here to stop the genservers calls of `Process.send_after(self(), {:decide_action, bot_id}, @decide_delay)` since even if it's being handled in the `decide_action/4` method it doesn't stop the calls since this is being called on each "decide_action" message. I'll avoid changing the bot_state to nil to keep consistency tho

AminArria (approved):
LGTM (but didn't live test it)

AngieDutra (dismissed):
Tested it and it works a lot better now!

AngieDutra (approved):
Tested it in mobile, tried a couple of variants and it worked great! Congrats ✨

PR #1009, opened 2023-10-20
Fix zone timer on reconnect

### LINKED ISSUE #818: [BUG] Zone time on reconnect
## 🐛 Bug Report

When you reconnect to a game , the zone timer starts again. This is because we are mapping the GameSettings.json `map_shrink_wait_ms` value. We need to track the current zone time of the game.

tkz00:
Registered the start time of the match, calculate based on that the time elapsed since the beginning of it, subtract from the time it takes to start shrinking the zone the time elapsed and display it.

Closes #818

### DIFF
--- a/client/Assets/Scripts/SocketConnectionManager.cs
+++ b/client/Assets/Scripts/SocketConnectionManager.cs
@@ -163,6 +163,11 @@ public class SocketConnectionManager : MonoBehaviour
                     this.gameProjectiles = gameEvent.Projectiles.ToList();
                     alivePlayers = gameEvent.Players.ToList().FindAll(el => el.Health > 0);
                     updatedLoots = gameEvent.Loots.ToList();
+
+                    if(LobbyConnection.Instance.matchStartTime == 0) {
+                        LobbyConnection.Instance.matchStartTime = (ulong)gameEvent.ServerTimestamp;
+                        PlayerPrefs.SetString("matchStartTime", ((ulong)gameEvent.ServerTimestamp).ToString());
+                    }
                     break;
                 case GameEventType.PingUpdate:
                     currentPing = (uint)gameEvent.Latency;
@@ -186,8 +191,6 @@ public class SocketConnectionManager : MonoBehaviour
                     );
                     this.allSelected = true;
                     this.gamePlayers = gameEvent.Players.ToList();
-                    LobbyConnection.Instance.matchStarted = DateTime.Now;
-                    PlayerPrefs.SetString("matchStartTime", DateTime.Now.ToString());
                     break;
                 default:
                     print("Message received is: " + gameEvent.Type);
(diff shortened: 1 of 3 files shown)


BertovDev (changes requested):
When timer reaches 0:

Captura de pantalla 2023-10-23 a la(s) 14 44 43

client/Assets/Scripts/UI/MatchStatsController.cs
BertovDev, line 1: LGTM! consider doing something like this to improve code readability:

```
        ulong mapShrinkWaitMs = LobbyConnection
            .Instance
            .serverSettings
            .RunnerConfig
            .MapShrinkWaitMs;
```

tkz00: Looks good! Already made the changes!

client/Assets/Scripts/UI/MatchStatsController.cs
```diff
@@ -39,13 +25,22 @@ void FixedUpdate()
...
+                .serverSettings
+                .RunnerConfig
+                .MapShrinkWaitMs;
+            if(mapShrinkWaitMs > elapsedTime) {
+                remainingSeconds = (mapShrinkWaitMs - elapsedTime) / 1000;
+                zoneTimer.text = remainingSeconds.ToString();
+            }
+            else {
+                remainingSeconds = 0;
+                zoneTimer.text = "0";
```
ncontinanza, line 42: In both cases, having `zoneTimer.text = remainingSeconds.ToString();` will do. What do you think of leaving that line only once outside de if/else? After that, maybe you can go further and rewrite the assignment as a single statement using ternary operation if you feel it more readable/cleaner.

ncontinanza (changes requested):
Working great! Just a little comment about the code

client/Assets/Scripts/SocketConnectionManager.cs
```diff
@@ -163,6 +163,11 @@ private void OnWebSocketMessage(byte[] data)
                     this.gameProjectiles = gameEvent.Projectiles.ToList();
                     alivePlayers = gameEvent.Players.ToList().FindAll(el => el.Health > 0);
                     updatedLoots = gameEvent.Loots.ToList();
+
+                    if(LobbyConnection.Instance.matchStartTime == 0) {
+                        LobbyConnection.Instance.matchStartTime = (ulong)gameEvent.ServerTimestamp;
+                        PlayerPrefs.SetString("matchStartTime", ((ulong)gameEvent.ServerTimestamp).ToString());
```
saugon, line 169: This approach to solving the issue could be appropriate only if it weren't possible to make any changes on the backend side.

In our case, the best approach is to obtain the current time directly from the backend and display it without requiring any calculations on the client side.

tkz00: That's right, it should be the backend's responsibility to calculate and return that information, I'll close this PR and we are going to fix the issue with the corresponding backend and client changes in a new PR.

tkz00:
After some consideration we came to the conclusion the best approach to display the remaining time for the damage zone would be to return the number from the backend, instead of calculating it in the front end, so we'll close this PR and fix the problem properly in a future one.

PR #1017, opened 2023-10-20
Make the bot target an entity in his proximity

agustinesco:
This pr is part of #709 

Curretly the bots that you spawn will move in his site randomnly and try to attack any enemy, the objective of this pr is make the bots go to any entity the have in an area around them
- If the target is a loot they should go and pick it
- If the target is an enemy they should go close and attack it 

To test this pr you should have a local server and connect to it with 
```
make start
```

and then go into any kind of match quick play or create a lobby, when the match starts you can spawn a bot with the "spawn bot" button

Screenshot 2023-10-26 at 11 37 32

The most important thing to notice in this pr is if the detection area of the bots is big enough


server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -4,11 +4,15 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
   alias DarkWorldsServer.Engine.Runner
+  alias LambdaGameEngine.MyrraEngine.Position
   alias LambdaGameEngine.MyrraEngine.RelativePosition
 
-  # This variable will dice how much time passes between bot decisions in milis
+  # This variable will decide how much time passes between bot decisions in milis
   @decide_delay_ms 500
 
+  # We'll decide the view range of a bot
+  @visibility_max_range 2000
```
Nico-Sanchez, line 14: Let's explain what this variable does and what's the unit of it in the comment line, just like the one above.

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -158,40 +151,38 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
-          Runner.play(game_pid, bot_id, %ActionOk{
-            action: :basic_attack,
-            value: %RelativePosition{x: x, y: y},
-            timestamp: nil
-          })
-        end
+  defp do_action(bot_id, game_pid, _players, %{
+         action: {:try_attack, %{type: :enemy, direction_to_entity: {x, y}} = direction_to_entity}
+       }) do
+    # TODO replace this 400 with a function that determines if any skill would hit the enemy
```
Nico-Sanchez, line 157: The idea is to see if the enemy is reachable? If so the bot will attack otherwise will get closer?
If that's the case maybe we can add it to the "todo" tag or as a comment above the function.

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -158,40 +151,38 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
+    else
+      Runner.play(game_pid, bot_id, %ActionOk{
+        action: :basic_attack,
+        value: %RelativePosition{x: x, y: y},
+        timestamp: nil
+      })
     end
   end
 
+  defp do_action(bot_id, game_pid, _players, %{
```
Nico-Sanchez, line 169: Is this for loots? Maybe a comment would be nice so other devs doesn't have to guess.

server/lib/dark_worlds_server/engine/bot_player.ex
```diff
@@ -200,27 +191,100 @@ defmodule DarkWorldsServer.Engine.BotPlayer do
...
+          true ->
+            Enum.min_by([hd(loots_distances), hd(players_distances)], fn entity -> entity.distance_to_entity end)
         end
     end
   end
 
-  def decide_objective(_, _), do: :nothing
+  defp get_closes_entity(_, _) do
+    %{}
+  end
```
Nico-Sanchez, line 263: 😛 
(Beware the typo is everywhere)
```suggestion
  defp get_closest_entity(_, _) do
    %{}
  end
```

Nico-Sanchez (changes requested):
Nice refactor and additions you've done here @agustinesco. Congrats 🎉 !
The bots seem more alive now. The random movement is still weird but I think its job for another PR.

Left some comments to improve the code readability.

agustinesco:
> Also, why is the bot test deleted?

I deleted them since they were unit test and they wasn't being helpful, we should only do integration testing

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: whoops, should be fixed now

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: Sure thing!

server/lib/dark_worlds_server/engine/bot_player.ex
agustinesco: Yes! this match is for the loot entities, comment added

Nico-Sanchez (approved):

AngieDutra (approved):
LGTM ✨

PR #1022, opened 2023-10-20
Fix bot spawning with dog model

agustinesco:
The bots will spawn with a random character defined in the config file now


client/Assets/Scenes/Battle.unity
BertovDev, line 1: You dont need to change the `Battle` scene. Change already did in the `CameraUI` prefab

`git checkout main -- client/Assets/Scenes/Battle.unity`

BertovDev (changes requested):

AminArria (changes requested):
Update `mix.lock` with `mix deps.get lamda_game_engine` once the other PR is merged

AminArria (approved):
LGTM (only looked at backend part)

client/Assets/Scenes/Battle.unity
agustinesco: Sure! i'm not sure what happened here

BertovDev (changes requested):
If dog prefab is deleted and not used anymore their assets should also have to be removed.

you can find them in `client/Assets/ThirdParty/DogKnight` I would delete the whole folder.

agustinesco:
> If dog prefab is deleted and not used anymore their assets should also have to be removed.
> 
> you can find them in `client/Assets/ThirdParty/DogKnight` I would delete the whole folder.

Sure thing, i deleted the folder and everything seems to be working correctly

AminArria (approved):

Issue #636, opened 2023-07-25
[FEATURE] KillFeed when killed by the damage zone


ncontinanza:
## 🚀 Feature

Currently we show the killfeed when a player kills another one, but not when a player dies by effect of the damage zone.

PR #1023, opened 2023-10-20
Fix kill cam

BertovDev:
This PR fixes the killcam. It uses the killfeed messages to find the correct player to follow.

To test:

- If you get killed by a player the camera should spectate that player.
- If you get killed by the danger zone you should spectate the first player alive in the list ( ex: player1 )

This PR patches the problem that the backend doesn't send any msg if you get killed by the danger zone, so are doing the following:

- Check if the player following is DEAD
- Check if there is no KilledBy event
- Result : you get killed by the dangerZone

DISCLAIMER:
`killedRegistrations.Add(9999);` I hate this line but it was necessary for the patch. I will implement the backend version detailed in this #636 and this will be deleted!

### DIFF
--- a/client/Assets/Scripts/CustomLevelManager.cs
+++ b/client/Assets/Scripts/CustomLevelManager.cs
@@ -373,26 +373,18 @@ public class CustomLevelManager : LevelManager
 
     private void SetCameraToAlivePlayer()
     {
-        playerToFollow = Utils.GetGamePlayer(KillFeedManager.instance.playerToTrack);
-        if (KillFeedManager.instance.saveKillerId != 0)
-        {
-            StartCoroutine(WaitToChangeCamera(playerToFollow));
-        }
-        else
-        {
-            if (playerToFollow.Status == Status.Dead && Utils.GetAlivePlayers().Count() > 0)
-            {
-                playerToFollow = Utils.GetAlivePlayers().ElementAt(0);
-                setCameraToPlayer(playerToFollow.Id);
-            }
-        }
+        playerToFollow =
+            KillFeedManager.instance.killedByDangerZone == true
+                ? playerToFollow = Utils.GetAlivePlayers().ElementAt(0)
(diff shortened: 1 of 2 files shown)


client/Assets/Scripts/CustomLevelManager.cs
```diff
@@ -373,12 +373,18 @@ private IEnumerator ShowDeathSplash(GameObject player)
...
         {
             StartCoroutine(WaitToChangeCamera(playerToFollow));
         }
-        else
+        if (
+            (
+                playerToFollow.Status == Status.Dead
+                && KillFeedManager.instance.saveKillerId == 0
+                && Utils.GetAlivePlayers().Count() > 0
+            )
```
Nico-Sanchez, line 386: Two things regarding these conditions:
- `playerToFollow` can be null? If so, it will break when you try to get the status in the 2nd if (`playerToFollow.Status`).
- Why are the 2 conditions not connected with each other? I believe they will be in conflict if both conditions are met. We should have one behavior for each scenario.

Also, the condition is wrapped in double parenthesis `(())`. One is enough.

Nico-Sanchez (changes requested):
Tested it with 3 devices and it works properly. However we should take a look at the code because I believe it will break or behave weirdly in some cases.

client/Assets/Scripts/CustomLevelManager.cs
BertovDev: You are totally right. I did some mistakes i will explain:

- `PlayeToFollow` can never be null. It is set in `KillFeedManager` with your player, then if you get killed it is set as the killer or if you get killed by the dangerZone it is set as the first alive player.
- Now i connect the 2 conditions.

I changed the code with your suggestions and implement a new logic. I believe its better now.

What do you think?

Nico-Sanchez: Nice one. Thanks for the explanation 😄 

As we discussed at the office, we'll leave a comment for a future refactor when the backend sends the "killed by zone" event.

PR #1026, opened 2023-10-23
Add 3D sound to looteables

### LINKED ISSUE #1016: [BUG] Boxes picked up far away from the player should no trigger the sfx
Only boxes picked up near the player should play the sfx, following the same logic as the skills sfx.

#### Reference: https://github.com/lambdaclass/curse_of_myrra/issues/812

tkz00:
Added "3DSoundManager" game object to LootBox prefab hierarchy, which contains the scripts necessary to play 3D audio and changed how the audio is played in Loot.cs for it to be 3D.

Unless Loot.cs logic for playing audio is changed, every looteable game object will need a "3DSoundManager" or similar GO in it's hierarchy since it's required to play sounds with the current logic. This could need refactoring if we really need a looteable to not play a 3D sound for some reason, but if we just need to make the sound be heard by all the players in the arena we can simply extend the range of the feedback sound.

Closes #1016


client/Assets/Scripts/Loot.cs
```diff
@@ -68,11 +68,8 @@ private void RemoveLoot(ulong id)
         GameObject lootObject = GetLoot(id).lootObject;
         string type = GetLoot(id).type;
 
-        MMSoundManagerSoundPlayEvent.Trigger(
-            GetLootableByType(type).pickUpSound,
-            MMSoundManager.MMSoundManagerTracks.Sfx,
-            lootObject.transform.position
-        );
+        lootObject.GetComponentInChildren<Sound3DManager>().SetSfxSound(GetLootableByType(type).pickUpSound);
```
saugon, line 71: Wouldn't it be better to add the script directly to the lootbox instead of searching through its children?

Also, instead of calling GetComponent (any variant), you could add a reference to the sound manager in Loot.cs. By doing this you will avoid multiple searches.

What do you think?

tkz00: Yes, both things you say are right, I built it like that copying how the skills and it's relation to the sound manager were coded, but in the case of skills it had to be that way since they are assigned dynamically on runtime to the character game object, which holds the sound manager (maybe this could also be improved, don't know), but in the case of looteables, since they are built as prefabs, we can construct them as you say, making it more organized and without the need to use any variant of GetComponent. I will correct this tomorrow.

tkz00: I'll start working on those changes right now. I have to create a new branch for the changes not related to the sound, right? And later reference the new branch in response to this messages so the conversation thread isn't lost.

tkz00: I've solved the 3D sound issue in this PP, reopened it for review, and created this new [issue](https://github.com/lambdaclass/curse_of_myrra/issues/1058) for the refactoring of the loot classes, in which I'll be working on next.

client/Assets/Scripts/Loot.cs
```diff
@@ -68,11 +68,9 @@ private void RemoveLoot(ulong id)
         GameObject lootObject = GetLoot(id).lootObject;
         string type = GetLoot(id).type;
 
-        MMSoundManagerSoundPlayEvent.Trigger(
-            GetLootableByType(type).pickUpSound,
-            MMSoundManager.MMSoundManagerTracks.Sfx,
-            lootObject.transform.position
-        );
+        var lootSoundManagerRef = lootObject.GetComponent<Sound3DManager>();
```
saugon, line 71: While using 'var' doesn't change anything at the compile/execution level, it would be better to use in general the expected type for the sake of readability.

tkz00: Right, already made the change!

saugon (approved):
Nice, good job!

ncontinanza (approved):
LGTM!

PR #1028, opened 2023-10-23
Fix death splash

### LINKED ISSUE #1025: [BUG] Death/win splash not being displayed
## 🐛 Bug Report
The match report should always be displayed, whether winning or losing.

BertovDev:
This PR fixes the Final death Splash.

The problem:
- When the match ends and you were in espectate mode. Final splash wat not showing correctly.

Be careful!
There is 2 death screen. One when you get kill and the final when the match ends. Make sure that the last one is working correctly. It should show you the winner player. If you win or lose.

closes #1025


client/Assets/Scripts/UI/DeathSplash/DeathSplashManager.cs
```diff
@@ -76,7 +76,7 @@ void OnEnable()
         ShowMessage();
         ShowMatchInfo();
         ShowPlayerAnimation();
-        ShowEndGameScreen();
+        // ShowEndGameScreen();
```
AngieDutra, line 79: If this is not being used, I would delete it

client/Assets/Scripts/UI/DeathSplash/DeathSplashManager.cs
```diff
@@ -174,30 +174,20 @@ private void ShowPlayerAnimation()
     public void ShowEndGameScreen()
     {
         // TODO: get image from lobby
-        if (SocketConnectionManager.Instance.GameHasEnded())
+        print(SocketConnectionManager.Instance.GameHasEnded());
```
AngieDutra, line 177: I believe you forgot a print here

AngieDutra (changes requested):

AngieDutra (dismissed):
Tested in mobile and working great!

client/Assets/Scripts/CustomLevelManager.cs
```diff
@@ -57,6 +57,8 @@ public class CustomLevelManager : LevelManager
 
     private bool deathSplashIsShown = false;
 
+    public bool test = false;
```
agustinesco, line 60: I think we should delete this

BertovDev: Done

ncontinanza (approved):
LGTM! Tested in Unity and Android

agustinesco (approved):
LGTM! the death splash seems to be working correctly, tested both unity and android versions

PR #1029, opened 2023-10-23
Integration tests

martinacantaro:
Test flow through the UI to create a lobby and start a game. 
To do: use Unity UI interaction library instead of calling scripts directly.

- How to test:
Open the test runner to list and be able to run the tests
image

In the test runner you have multiple options to run tests separately, all together or just run the ones that failed last time you ran them
image

The "IsServerHealthy" test checks if the server is working fine, if it fails, the other tests that depend on the state of the server will too, those are "UnmuteButton", "BacktoLobbiesButton" and "MatchStart".

Recently we added a functionality to allow the player to choose his or her name at the start of the application (it's saved and will not ask for it again unless deleted), this name is necessary to create or join a match, and the test's flow don't contemplate the name functionality yet, so before running the tests it's needed to normally start the application and enter a name, then the tests will use that player name, if this is not done the tests that rely on having a name will fail. This will be fixed in the near future.


ncontinanza:
We noticed several auto-generated files such as client/Assets/InitTestScene638336713701830240.unity which shouldn't exist after running the whole suite of tests (https://forum.unity.com/threads/scene-testing.575920/)

martinacantaro:
> We noticed several auto-generated files such as client/Assets/InitTestScene638336713701830240.unity which shouldn't exist after running the whole suite of tests (https://forum.unity.com/threads/scene-testing.575920/)

Ready for review again!
- Those files are generated by running a test
- Sometimes they are deleted automatically, sometimes not. 
- In particular, they are not deleted automatically if you stop the test before it's completed
- We explored the solution of deleting these files in the test TearDown, but if you manually stop the test before completion, the TearDown would not run and that solution would not suffice
- Therefore, we added these files to the gitignore

.gitignore
```diff
@@ -66,6 +66,10 @@ ExportedObj/
 *.pdb.meta
 *.mdb.meta
 
+# Unity Test Framework generated files on interrupted tests
+client/Assets/*.unity
```
AngieDutra, line 70: I haven't checked yet but I believe this will difficult developers when creating new scenes that we know are coming up. I would be specific. We know the file created always starts with InitTest, for example

AngieDutra (changes requested):
Captura de pantalla 2023-11-01 a la(s) 16 37 56

I believe all test should pass so maybe recent merges broke some of them. This is something to check

.gitignore
AngieDutra: It turns out this is not an issue

tkz00: I've made a little change just for clarity.

tkz00:
Further investigation is needed to understand why certain tests fail, particularly those related to the server connection. When the "ServerIsHealthy" test fails, it will undoubtedly lead to the failure of many other tests. However, it's crucial for the "ServerIsHealthy" test to pass because its success is essential for obtaining accurate results from the rest of the tests. If "ServerIsHealthy" fails do not take into account the rest of the tests results and rerun them until "ServerIsHealthy" passes.

client/Assets/Scripts/UnityTests/PlayModeTests/ServerHealthCheck.cs
```diff
@@ -0,0 +1,52 @@
...
+        yield return CoClickButton(launchGameButton);
+        yield return new WaitForSeconds(2f);
+
+        Assert.That(SceneManager.GetActiveScene().name, Is.EqualTo("CharacterSelection"));
+        yield return new WaitForFixedUpdate();
+
+        Assert.That(SocketConnectionManager.Instance.isConnectionOpen, Is.EqualTo(true));
+    }
+
+    IEnumerator CoClickButton(GameObject go)
```
AngieDutra, line 41: I saw this in TestingUtils. Maybe use it here too?

AngieDutra (changes requested):

client/Assets/Scripts/UnityTests/PlayModeTests/ServerHealthCheck.cs
```diff
@@ -0,0 +1,52 @@
...
+        base.Setup();
+        SceneManager.LoadScene("Scenes/Lobbies");
+    }
+
+    // Creates a lobby in the Lobby scene
+    [UnityTest]
+    public IEnumerator ServerIsHealthy()
+    {
+
+        GameObject newLobbyButton = GameObject.Find("NewLobbyButton");
```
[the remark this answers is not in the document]
tkz00: You're right! I'll add the search for the element to click to the click method instead of repeating that line every time.

AngieDutra (changes requested):

client/Assets/Scripts/UnityTests/PlayModeTests/LobbiesUITest.cs
```diff
@@ -0,0 +1,58 @@
...
+        yield return new WaitForSeconds(5f);
+        Assert.That(SceneManager.GetActiveScene().name, Is.EqualTo("Lobby"));
+
+        yield return TestingUtils.ForceClick("LaunchGameButton");
+        yield return new WaitForSeconds(5f);
+        Assert.That(SceneManager.GetActiveScene().name, Is.EqualTo("CharacterSelection"));
+
+        yield return TestingUtils.ForceClick("Back");
+        yield return new WaitForSeconds(5f);
+        //Assert.That(SceneManager.GetActiveScene().name, Is.EqualTo("Lobbies"));
```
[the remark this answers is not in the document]
tkz00: Uncommented the line!

client/Assets/Scripts/UnityTests/PlayModeTests/ServerHealthCheck.cs
AngieDutra, line 1: I have a theory on how to make this tests better. The error when this test fails is the following:

Captura de pantalla 2023-11-07 a la(s) 13 31 45
This indicates the socket check is failing when there is no instance of the player. We could control the check by running it when the player is available in the socket instance. This is an idea for a future implementation.

AngieDutra: Maybe in issue #1020

AngieDutra (dismissed):
LGTM ✨

ncontinanza:
I got this error when trying it out in my client:

```
MatchStart (13.971s)
---
Expected string length 6 but was 18. Strings differ at index 0.
  Expected: "Battle"
  But was:  "CharacterSelection"
  -----------^
---
at MatchStartFlow+<MatchStart>d__1.MoveNext () [0x00227] in /Users/ncontinanza/Projects/Lambdaclass/curse_of_myrra/client/Assets/Scripts/UnityTests/PlayModeTests/MatchStartFlow.cs:42
at UnityEngine.TestTools.TestEnumerator+<Execute>d__7.MoveNext () [0x0003a] in /Users/ncontinanza/Projects/Lambdaclass/curse_of_myrra/client/Library/PackageCache/com.unity.test-framework@1.1.33/UnityEngine.TestRunner/NUnitExtensions/Attributes/TestEnumerator.cs:46

```

Am I missing something?

ncontinanza (dismissed):

ncontinanza (approved):

AngieDutra (approved):
Nice work!

ncontinanza (approved):

Issue #885, opened 2023-09-12
[BUG] Left joystick has a fixed max range


AngieDutra:
This is an issue the max range doesn't adapt to each resolution and makes de usability different for each resolution

PR #1031, opened 2023-10-23
Responsive max range

AngieDutra:
This PR closes issue #885 

## To test this simulate different devices in the simulator tab of Unity.
Captura de pantalla 2023-10-23 a la(s) 19 05 25

Keep in mind for every test is each simulation screen you should restart the play mode.

### Check:

- if the knob's max range is visually the same in each simulation. 
(It should be different in each simulation numbers wise since it is know responsive and adapts to each screen)

- When you press each border the joystick's knob and background stay within the screen limits and are easy to use

It's important each reviewer tests this in a mobile build too


BertovDev (approved):
- [x] Tested in Unity Desktop
- [x] Tested in Android

agustinesco (approved):
LGTM! 

tested both on android build and with various simulators and seems to be working as expected

Issue #965, opened 2023-10-10
[DESIGN] New buttons layout


saugon:
## 🎨 Design
Characters will have 3 possible buttons/actions now:
- Basic skill: would be the most used button and always available.
- Special skill: starts as disabled and depending on different actions of the player, it changes to enabled in order to be used. Once used, it returns to its previous state.
- Item: at first it will be empty (therefore disabled/hidden) until the player picks up an item, displaying inside the button that item. Once used, it returns to its previous state.

PR #1037, opened 2023-10-24
New buttons layout

AngieDutra:
This PR closes issue #965 

Now we should have only 2 skills and one special attack.
super-disabled

It's important to test the following items in a mobile build:
- if the buttons feel the right size 
- if the spacing between them feels right or if it makes it harder for the user to reach one


agustinesco (approved):
LGTM!
- Tested the buttons on a mobile build and it seems good and they're comfortable to use
- Tested on the unity simulator with a few resolutions and they seem fine

BertovDev (approved):

PR #1043, opened 2023-10-24
EngineRunner support Muflus hit skill mechanic

AminArria:
Related to https://github.com/lambdaclass/lambda_game_engine/pull/22

This update implemetents the following
- Hit skill

Pending
- [x] Update `mix.lock` once merged https://github.com/lambdaclass/lambda_game_engine/pull/17

## Testing this PR

In `config.exs` change the config to
```
config :dark_worlds_server, DarkWorldsServer.Engine.Runner,
  process_priority: :high,
  use_engine_runner: true  ## <- Setting true enables EngineRunner to replace Runner
```

You can select Muflus for testing the hits, currenly only the basic attack and skill 1 (lowest circle) have an associated skill
- Basic attack: Hit in a 500 range in a 90 degree cone
- Skill 1: Hit in a 500 range in a 360 degree 

**Note:** There is not going to be any animation on the client side as there is no action been sent from the server

Bots can be manually spawned if you want, you will need the `EngineRunner` PID which can be seen in the log line

```
[info] Switching to engine_runner #PID<0.642.0>
```

And then you only need to invoke this helper function with the middle number of the PID

```
iex(21)> DarkWorldsServer.Engine.EngineRunner.bot_join(642)
```


manucamejo (dismissed):
Nice work.
I found two things we might not need to worry about:
- When I spawn a bot (with the shell), it doesn't appear unless I select the dev tab.
- If I press `spawn bot`, the server crashes.

manucamejo (approved):

Nico-Sanchez (approved):
Nice job!

PR #1055, opened 2023-10-25
Fix Danger zone UI

### LINKED ISSUE #1047: [BUG] Safe zone UI feedback not working
## 🐛 Bug Report
The UI feedback when exiting the safe zone is not functioning.

BertovDev:
The problem:
- The Danger zone UI sometimes was not appearing correctly.

What happened ?
- When someone die a method called `SetPlayerDead` in `Battle.cs` called  `CustomGUIManager.DisplayZoneDamageFeedback(false);` causing the bug

The Fix:

- We only want to desactivate the zone effects if we are dead and it is in our client, so:
	```
	if (SocketConnectionManager.Instance.playerId == ulong.Parse(playerCharacter.PlayerID))
	    {
	            CustomGUIManager.DisplayZoneDamageFeedback(false);
	     }
	```

closes #1047


agustinesco (approved):
LGTM!

ncontinanza (approved):

PR #1057, opened 2023-10-25
Fix bots animation

manucamejo:
Problem:
Bots don't have animations because they are created in the middle of the game. So, when they spawn, they don't have related skills in Unity.

Solution:
Adding the skills to the bot when spawned.

### DIFF
--- a/client/Assets/Scripts/CustomLevelManager.cs
+++ b/client/Assets/Scripts/CustomLevelManager.cs
@@ -204,6 +204,50 @@ public class CustomLevelManager : LevelManager
 
                     healthBarFront.color = Utils.healthBarRed;
                     SocketConnectionManager.Instance.players.Add(newPlayer.gameObject);
+
+                    List<Skill> skillList = new List<Skill>();
+
+                    SkillBasic skillBasic = newPlayer.gameObject.AddComponent<SkillBasic>();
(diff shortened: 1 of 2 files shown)


client/Assets/Scripts/CustomLevelManager.cs
```diff
@@ -204,6 +204,50 @@ private void GenerateBotPlayer(SocketConnectionManager.BotSpawnEventData botSpaw
...
+                    string selectedCharacter = SocketConnectionManager.Instance.selectedCharacters[
+                        UInt64.Parse(newPlayer.PlayerID)
+                    ];
+                    CoMCharacter characterInfo = charactersInfo.Find(
+                        el => el.name == selectedCharacter
+                    );
+                    SkillAnimationEvents skillsAnimationEvent =
+                        newPlayer.CharacterModel.GetComponent<SkillAnimationEvents>();
+
+                    List<SkillInfo> skillInfoClone = InitSkills(characterInfo);
```
saugon, line 225: This line and the following one are not necessary (because bots don't use the UI)

saugon: Never mind, this improvement requires additional changes that will fall outside the scope of this fix.

saugon (approved):

BertovDev (approved):

PR #1060, opened 2023-10-26
Fix death splash broken on tests

BertovDev:
Thanks to the integrations test we could find a problem in the flows of the screen.

The problem:

- `Battle` scene was been pre-loaded in the `characterSelection` Scene and `DeathSplashManager` was active and searching for non existing variables in that moment causing erorrs in the tests.

The solution:

- Desactivate `deathSplash` prefab as default and activating it when needed.


AngieDutra (approved):
IMO we could change when the death splash requests data. Why would it be okey for the death splash to requiere data when the used id selecting a character? But setting the active state of the game object to false when starting the scene looks like and easy fix for now. I'm interested in other opinions on this subject.

AngieDutra (approved):

ncontinanza (approved):
LGTM!

Issue #926, opened 2023-10-03
[FEATURE] Character list screen


saugon:
Implement character list design

PR #1061, opened 2023-10-26
Character list

AngieDutra:
This PR closes issue #926 
Refe-1

Check:
- you can go from MainScreen to this screen and back to Main
- the list scrolls as expected
- the information is right
- the pressed state reduces the element's scale

The only files with code to check are:
- client/Assets/Scripts/CoMCharacter.cs
- client/Assets/Scripts/UI/ButtonAnimations.cs
- client/Assets/Scripts/UI/CharactersListManager.cs

Note: this PR changes Lobbies.unity file due to UI problems solved when updating the back button arrow


client/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -167,4 +167,7 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/Scenes/TitleScreen.unity
     guid: 4900ba75a16aa4edbb73129a2a725d34
+  - enabled: 0
```
Nico-Sanchez, line 170: I had an error with the character list scene and had to enable it from Unity.

I believe if you enable it here that error will be fixed. Can we do that or this is zero on purpose?

client/Assets/Scripts/UI/CharactersListManager.cs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    void GenerateList()
+    {
+        players.ForEach(player =>
+        {
+            GameObject item = (GameObject)Instantiate(listItem, this.transform);
+            item.GetComponentInChildren<Image>().sprite = player.characterSprite;
+            item.GetComponentInChildren<TextMeshProUGUI>().text = player.name;
+            //item.GetComponentInChildren<TextMeshPro>().text = player.name;
```
Nico-Sanchez, line 28: Can we remove this comment? Do we want to keep this?

Nico-Sanchez (changes requested):
The character list scene looks good. Nice job!

Left some comments regarding the code and an error I had when testing it.

client/Assets/Scripts/UI/CharactersListManager.cs
AngieDutra: On it!

Nico-Sanchez (dismissed):

client/Assets/Scripts/UI/CharactersListManager.cs
```diff
@@ -0,0 +1,30 @@
...
+        GenerateList();
+    }
+
+    void GenerateList()
+    {
+        players.ForEach(player =>
+        {
+            GameObject item = (GameObject)Instantiate(listItem, this.transform);
+            item.GetComponentInChildren<Image>().sprite = player.characterSprite;
+            item.GetComponentInChildren<TextMeshProUGUI>().text = player.name;
```
BertovDev, line 27:
```suggestion
        CharactersScriptables.ForEach(character =>
        {
            GameObject item = Instantiate(listItem, this.transform);
            item.GetComponentInChildren<Image>().sprite = character.characterSprite;
            item.GetComponentInChildren<TextMeshProUGUI>().text = character.name;
```

BertovDev (changes requested):

client/Assets/Prefabs/UI/CharacterListItem.prefab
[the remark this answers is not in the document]
AngieDutra: Talked with our TL and agreed it wasn't wrong to do it in order to solve design issues. With that being said design changes were made and functionality changes resulted in this not being an issue anymore

client/Assets/Scripts/UI/CharactersListManager.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+using TMPro;
+
+public class CharactersListManager : MonoBehaviour
+{
+    [SerializeField]
+    List<CoMCharacter> players;
```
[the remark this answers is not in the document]
AngieDutra: I agree on this but changed it to CharacterScriptableObjects

client/Assets/Scripts/UI/CharactersListManager.cs
AngieDutra: This was already adressed

BertovDev (changes requested):
Characters Button in mainScreen is not working.

AngieDutra:
> Characters Button in mainScreen is not working.

Are you sure? I'm testing it and it works just fine to me. Make sure to update your branch and to close and open the Unity project

BertovDev (dismissed):

tvillegas98 (approved):
LGTM!

AngieDutra:
> I dont quite understand the `IsBackButton` functionality in `ButtonAnimationsMMTouchButton` can you please bring a little more explanation about it.

Sure! isBackButton is useful to identify back buttons. Although these should have the same looking scale animation due to its different size the scale difference should be bigger. It is also helpful to identify the buttons that trigger a change of scene to avoid the error of animating an element that is no longer there. I believe there is room for improvement tho, but this was the first implementation

BertovDev:
> > I dont quite understand the `IsBackButton` functionality in `ButtonAnimationsMMTouchButton` can you please bring a little more explanation about it.
> 
> Sure! isBackButton is useful to identify back buttons. Although these should have the same looking scale animation, due to its different size the scale difference should be bigger to be noticeable. It is also helpful to identify the buttons that trigger a change of scene to avoid the error of animating an element that is no longer there. I believe there is room for improvement tho, but this was the first implementation

Nice!, thanks for the explanation

BertovDev (approved):
Nice!

PR #1063, opened 2023-10-26
UI Bug fixes

AngieDutra:
This PR fixes bugs the team noticed during one play test. Check:

- Exit buttons does no longer overlay with the settings screen. To test this create a match, die but don't be the last one to die. I tried this with bots. Enter Spectate mode and open the settings screen. 
- The sprite needed in title screen was restored
- KillFeed text was changed to uppercase and the font was also changed to the proper one


Nico-Sanchez (approved):
Let's gooooo 🚀

PR #1067, opened 2023-10-27
Add bot wandering

agustinesco:
This pr is part of #709 

The bots should pick a random location in the playable area (inside the zone) and walk to it now, they can get distracted if they come across a player of loot crate on their way

To pick their next wander position the posible destinations should meet the next requirements:
- Be inside of the bot visibility zone
- Be inside of the board
- Be inside of the safe zone

i'll attach some examples of possible destinations:
Screenshot 2023-10-26 at 12 51 56
Screenshot 2023-10-26 at 12 52 44
Screenshot 2023-10-26 at 12 53 13


martinacantaro (approved):
LGTTTMMMMMM!!! 🚀 
Playtested and it looks good
Went through the code with Agus and from what I see it makes sense

martinacantaro (approved):

ncontinanza (approved):

PR #1068, opened 2023-10-27
Make bot attack only if his basic skill reaches

agustinesco:
This pr is part of #709 

The aim of this pr is to make the spawned bots to use their basic skill if it would hit an enemy.

For the time being we'll just use the skill.json file to determine if the basic skill reach an enemy, when the new engine is implemented we should update this to do a propper check

The check doesn't need to be fully accurate, we should add a field to the skills that determines the reach of the skill, something like [short, mid, long] and use that to determine if the basic attack hits


axlrott (approved):
LGTM! Good job @agustinesco

manucamejo (approved):

PR #1069, opened 2023-10-27
Make bots take random decissions

agustinesco:
This pr is part of #709 

We'll make the bots behave randomly make them sometimes walks instead of attack and in a lower rate to just stand still

Also we'll add some inaccuracy to the bots hits so they'll miss sometimes


axlrott:
This PR has an old version of the logic of the bots wich generate conflicts that are not easy to solve.
I'll create another PR from master with this logic added.

PR #1084, opened 2023-11-01
Add more behaviours to bots and some refactoring

axlrott:
This PR is part of [#709](https://github.com/lambdaclass/curse_of_myrra/issues/709)

## Motivation
Bots were always attacking with a perfect precision, they never changed their behaviour even when they had low health and also sometimes they were on the edge or entering the unplayable area wich lowered their health.

## Summary of changes
* Made some refactors in the code to make it a little bit more clear, mostly on functions where we could use some pattern matching 
* Divide the objective `:chase_entity` to `:chase_loot` and `:chase_enemy` to separate the different behaviours
* Change logic to avoid the not playable zone, now the bot will avoid that zone and a little bit more of space so we can't be on the edge of both zones, this could sometimes result in losing health because the bot went to the unplayable zone.
* Change the function `calculate_circle_point` so we can set if we want the returning direction  to have some inaccuracy, this will be use on the attacks.
* Add to the condition that if the bot has low health it should run away from players unless the playable radius is too low

## How has this been tested?
* The attacks can be tested mostly on h4ck, as it's the only character with a long range atack, by letting him attack you and see that sometimes the attack misses
* It can be tested looking at bots when they have low health that they run away, also when the playable radius is low they go back to attack
* Now the bots avoid the unplayable zone with a little margin so they should always stay on the playable zone and don't have weird behaviour

## Test additions / changes
No tests added

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


axlrott:
Close this PR as this logic is already implemented for the new flow -> https://github.com/lambdaclass/curse_of_myrra/pull/1104

PR #1085, opened 2023-11-01
Gh 998 item vfx

### LINKED ISSUE #998: [FEATURE] Used item vfx
## 🚀 Feature
Create a generic vfx to display when an item is used.

martinacantaro:
Closes #998

## Motivation

In the future, players will receive items from crates. When they use a item, an animation will be shown. The backend implementation is not yet done, so all we can do for now is add a vfx to use in the future.
- This is a single generic animation that will show for all 5 types of items
- Each type of item will have its own vfx on top of that in the future

## Summary of changes

Inside each character prefab, inside the FeedbackContainer, I added a UseItemEffect. 
This vfx is not called by any event currently, this PR is just to store it until we use it in the future.

## How has this been tested?

Entered playmode with the VFX added to the characterbase so that it's always showing, just to see how it looks since right now there's no event to trigger it.

## Test additions / changes
No tests were added or updated in this PR

## Checklist
- [-] Tests have been added/updated.
- [-] This change requires new documentation.
  - [-] Documentation has been added/updated.
- [-] I have tested the changes in another devices.
  - [-] Tested in iOS.
  - [-] Tested in Android.


BertovDev (changes requested):
Effect looking clean! and just few things to remark:

1. You dont need to change each character prefab. Only changing the FeedbackContainer.prefab as you did will be changed it each character prefab. So in order to restore you can do:
```
	git checkout main -- client/Assets/Prefabs/Characters/
```
2. I have notice that the effect in Muflus is a little bit more complicated to notice, maybe it is my perspective.  I leave a video below.


https://github.com/lambdaclass/curse_of_myrra/assets/62818001/03950e11-ab86-40d4-9aa9-0b710320c20c

What do you think?

BertovDev (commented):
So nice! Muflus feedback looking better now and now  that I re-read the PR description and realize you said 

> Each type of item will have its own vfx on top of that in the future

So I believe it will be useful to have the VFX reference in the ScriptableObject `Lootable.cs` for now just as a placeholder but it can be useful in the future implementation.

What do you think ?

BertovDev (approved):
Nice!

client/Assets/Scripts/Lootable.cs
```diff
@@ -7,6 +7,7 @@ public class Lootable : ScriptableObject
 {
     public string lootName;
     public GameObject lootPrefab;
+    public GameObject useItemVFX;
```
saugon, line 10: Does this attribute exist for a particular reason? If not, it should be removed.

martinacantaro: done!

BertovDev (approved):

saugon (approved):
Great job! 🤘🏻

PR #1098, opened 2023-11-03
End game (one player remains) and safeguard to kill process if zombie

AminArria:
- Checks game for a win condition of one (or none) player remaining, if it is the case triggers the game ending sequence
- Adds safeguard to kill lingering processes after 10 minutes

### DIFF
--- a/server/lib/dark_worlds_server/engine/engine_runner.ex
+++ b/server/lib/dark_worlds_server/engine/engine_runner.ex
@@ -9,6 +9,13 @@ defmodule DarkWorldsServer.Engine.EngineRunner do
   @game_tick_rate_ms 20
   # Amount of time between loot spawn
   @loot_spawn_rate_ms 20_000
+  ## Time between checking that a game has ended
+  @check_game_ended_interval_ms 1_000
+  ## Time to wait between a game ended detected and shutting down this process
+  @game_ended_shutdown_wait_ms 10_000
+  ## Timeout to stop game process, this is a safeguard in case the process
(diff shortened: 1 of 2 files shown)


server/lib/dark_worlds_server/engine/engine_runner.ex
```diff
@@ -160,6 +197,23 @@ defmodule DarkWorldsServer.Engine.EngineRunner do
...
+    myrra_state = transform_state_to_myrra_state(game_state)
+    Phoenix.PubSub.broadcast(DarkWorldsServer.PubSub, topic, {:game_ended, myrra_winner, myrra_state})
+  end
+
+  defp check_game_ended(players) do
+    players_alive = Enum.filter(players, fn player -> player.status == :alive end)
+
+    case players_alive do
+      ^players -> :ongoing
+      [_, _ | _] -> :ongoing
```
axlrott, line 211: Is the condition `^players -> :ongoing` necessary? I think that checking that there are more than one player alive is all we needed to say that the match is ongoing

axlrott (changes requested):

server/lib/dark_worlds_server/engine/engine_runner.ex
AminArria: You are correct, I think I left from a first attempt at writing this

AminArria: Removed in 1ce8af0e

axlrott (approved):
LGTM! 👍

PR #1139, opened 2023-11-14
Gh 1131 test player movement

### LINKED ISSUE #1131: [TESTS] Player movement test
## 🚧 Testing
Add an integration test with the backend, initializing a match, and verifying the player movement. 
Rather than attempting to use the UI input, simulate the input result, providing necessary values for the test, and trigger the input listener function from within the test.

tkz00:
Closes #1131 

## Motivation

It's necessary to test if the movement of the character works, if the client's input is received by the server and how it communicates the change of position back.

## Summary of changes

Created a test script that initializes a match, stores the player's position, then it programmatically moves the joystick top and right for half a second, returns it to it's idle state and compares the current position of the player with the one stored earlier, verifying it's X (horizontal) and Z (vertical) values are greater than the previous one's.

## How has this been tested?

Run the "PlayerMovement" test and check it passes.

## Test additions / changes

This PR is in itself a test.

## Checklist
- [X] Tests have been added/updated. (this is a test)
- [ ] I have tested the changes in another devices. (doesn't apply)
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


client/Assets/Scripts/UnityTests/PlayModeTests/BattleTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    // Creates a lobby in the Lobby scene
+    [UnityTest]
+    public IEnumerator PlayerMovement()
+    {
+        // Change to localhost server
+        yield return TestingUtils.ForceClick("ServerNameContainer");
+        yield return new WaitForSeconds(.1f);
+        yield return TestingUtils.ForceClick("LocalHost");
+        yield return new WaitForSeconds(.1f);
+
```
AngieDutra, line 25: I've noticed the screen view is blocked with warning and error popups. I recommend canceling them with the following lines:
`yield return TestingUtils.ForceClick("Ok");`
`yield return TestingUtils.ForceClick("No");`
We could ad a conditional to check first if they exist but I tried them in this file and they worked just fine

client/Assets/Scripts/UnityTests/PlayModeTests/BattleTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        yield return TestingUtils.ForceClick("ConfirmButton");
+        yield return new WaitForSeconds(1f);
+        Assert.That(SceneManager.GetActiveScene().name, Is.EqualTo("Battle"));
+        yield return new WaitForSeconds(3f);
+
+        #region Move Player
+        
+        // Get character initial position
+        GameObject player = Utils.GetPlayer(1);
+        Vector3 initialPlayerPosition = player.transform.position;
```
AngieDutra, line 52: Here I believe is clearer to replace the game object value with a Vector2 such as:
```
Vector2 leftJoystickMovementValue = GameObject.Find("JoystickKnob").GetComponent<LeftMMTouchJoystick>().RawValue;
        leftJoystickMovementValue = new Vector2(1, 1);
        yield return new WaitForSeconds(.5f);
        leftJoystickMovementValue = new Vector2(0, 0);
```

AngieDutra (changes requested):

client/Assets/Scripts/UnityTests/PlayModeTests/BattleTests.cs
tkz00: Already made the changes!

client/Assets/Scripts/UnityTests/PlayModeTests/BattleTests.cs
tkz00: You're right! I modified it to avoid calling GetComponent() multiple times!

AngieDutra (approved):
LGTM! If you have the time it would be great to add the logic to avoid popups during tests in all the current tests. But I believe the scope of this issue was achieved ✨

ncontinanza:
We'll leave this PR as draft until we can properly run the tests. Currently there are a couple of errors at the beginning of a match that are preventing the test from passing.

PR #1154, opened 2023-11-16
Main screen updates

### LINKED ISSUE #1042: [FEATURE] Main screen update
## 🚀 Feature
Implement the last version of main screen

AngieDutra:
Closes #1042 

## Motivation

A change of style was needed for UI-UX porpouses

## Summary of changes
reference

## Attention
Do not check Muflus's position into account when reviewing since this character is being refactor and this will be fix with that.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


manucamejo:
@AngieDutra I've 2 questions:
If I change the server, the "box" fits the word, is that correct? Or should we have the box width for all the different server names?
image
image


Also, the UI is different than the PR description, is that ok?
image

AngieDutra:
> @AngieDutra I've 2 questions: If I change the server, the "box" fits the word, is that correct? Or should we have the box width for all the different server names? image image
> 
> Also, the UI is different than the PR description, is that ok? image

The select button for the server is supposed to work that way. It adjusts to the server name.
The UI differences, if you refer to the transparent buttons, is because they are currently disabled. When we develop them I will enable them and the opacity will be like the image reference. If the differences you talked about were not about that, please tell me so

manucamejo (dismissed):

tkz00 (approved):
LGMT!

client/Assets/Scenes/MainScreen.unity
BertovDev, line 1: Do we need this changes in this scene?

AngieDutra: Yes we do because in main there are changes applied to the scene. Now it is all based on the prefab changes

martinacantaro:
A small obvservation: the `Play` button has a solid pink background, while the `daily quest`, `characters` and `news` buttons have a pink-black gradient.  I would add the gradient to the Play button as well, so that they all look coherent

client/Assets/Scripts/UI/ButtonAnimationsMMTouchButton.cs
```diff
@@ -5,45 +5,41 @@
 
 public class ButtonAnimationsMMTouchButton : MMTouchButton
 {
+    [Header("Back Button bool")]
+    [Tooltip(
+        "If this component is applied to a back button this should be truw to avoid animation errors"
```
ncontinanza, line 10:
```suggestion
        "If this component is applied to a back button this should be true to avoid animation errors"
```

ncontinanza (changes requested):

AngieDutra:
> A small obvservation: the `Play` button has a solid pink background, while the `daily quest`, `characters` and `news` buttons have a pink-black gradient. I would add the gradient to the Play button as well, so that they all look coherent

The button has a gradient between two similar magenta colors. Never the less, I spoke with the design team and they said a refactor on that is coming soon

BertovDev (approved):
Looks amazing!

ncontinanza (approved):

PR #1162, opened 2023-11-20
More metrics

AminArria:
## Motivation

We need more observability when running the load tests

## Summary of changes

- `new_relic_agent` dependency is fetched for all environments, but disabled in `config.exs` and only enable in `config/prod.exs`
- Add metrics for
  - Total player count
  - Total game count
  - Total websockets for games


mFragaBA (changes requested):
I think we should also update the `bot_count` from the state on `handle_info({:spawn_bots, bot_count}, state)` at `engine_runner.ex`, since we're changing the amount of players in the game

server/lib/dark_worlds_server/engine/engine_runner.ex
```diff
@@ -167,6 +170,8 @@ defmodule DarkWorldsServer.Engine.EngineRunner do
     now = System.monotonic_time(:millisecond)
     time_diff = now - state.last_game_tick_at
     game_state = LambdaGameEngine.game_tick(state.game_state, time_diff)
+    now_after_tick = System.monotonic_time(:millisecond)
+    NewRelic.report_custom_metric("GameBackend/GameTickExecutionTime", now_after_tick - now)
```
mFragaBA, line 174: oh, one small thing. Can we call this `GameTickExecutionTimeMs` to make the unit clear in newrelic?

mFragaBA, line 174:
```suggestion
    NewRelic.report_custom_metric("GameBackend/GameTickExecutionTimeMs", now_after_tick - now)
```

server/config/config.exs
```diff
@@ -78,6 +78,9 @@ config :dark_worlds_server, DarkWorldsServer.Engine.Runner, process_priority: :h
 hash = String.trim(hash)
 config :dark_worlds_server, :information, version_hash: hash
 
+# By default disable NewRelic agent
+config :new_relic_agent, license_key: nil
```
fkrause98, line 82: This line makes me raise an eyebrow, can we test that we're still reporting metrics in prod/load testing servers?.

mFragaBA (commented):
perhaps we should document somewhere that in order to see the values of `TotalGames`, `TotalGamesWebSockets` and `TotalPlayers` in the metrics explorer you need to use the `count` option

image

What do you think?

server/config/config.exs
mFragaBA: it is doing so. You can check on the newrelic board and also it's [custom metrics](https://one.newrelic.com/nr1-core/data-explorer/metrics/MzkyNzY2M3xBUE18QVBQTElDQVRJT058MTM3MjcyODQxMQ?account=3927663&duration=3600000&state=bb2984f2-3f1b-a5c8-7c99-4b46d9a8b1cf)

AminArria: This is from their [docs](https://github.com/newrelic/elixir_agent/tree/v1.27.8#disabling)

> If you want to disable the agent, you can do it in two different ways:
>
>Application config: config :new_relic_agent, license_key: nil
>Environment variables: NEW_RELIC_HARVEST_ENABLED=false

server/lib/dark_worlds_server/engine/engine_runner.ex
AminArria: Done in 8f95664

mFragaBA (approved):
LGTM!

fkrause98 (approved):
LGTM!

mFragaBA (approved):
LGTM!

mFragaBA:
Let's wait before merging since @manucamejo wanted to review this

PR #1165, opened 2023-11-21
Haptic feedback

### LINKED ISSUE #1111: [FEATURE] Haptic feedback
Add a small vibration feedback for the following events:
- When a player deals or receives damage.
- When a player cancels a skill.

AngieDutra:
Closes #1111 

## Motivation

The game is currently missing vibration feedback

## Summary of changes

Haptic feedback was added when the skill is canceled and when a player receives damage. To test this the reviewer must build this and test it on their phone

## Attention
I'm aware the issue involves applying the feedback for when it damages other players but we don't have that data from backend yet. For now we only know when the damage is applied to us but not who we attack. When that information is available the vibration can be added.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/CharacterFeedbacks.cs
```diff
@@ -90,6 +88,18 @@ public void ChangePlayerTextureOnDamage(float clientHealth, float playerHealth)
         }
     }
 
+    public void HapticFeedbackOnDamage(float clientHealth, float playerHealth, ulong playerId)
+    {
+        if (
+            clientHealth != playerHealth
```
tkz00, line 94: This condition is redundant with the previous one, the one on line 94 should be removed.

tkz00 (changes requested):

client/Assets/Scripts/CharacterFeedbacks.cs
AngieDutra: I agree. Good catch. Done ✨

tkz00 (approved):
Looks good!

BertovDev:
Nice! Working so good. It is possible to add more vibration when you are in the danger zone ? So it will feel like you are trully in danger.

AngieDutra:
> It is possible to add more vibration when you are in the danger zone ?

Sure! I believe it should be easy enough and it is within the scope of the issue since it involves damage feedback

PR #1189, opened 2023-11-23
Gh 1181 hide characters

### LINKED ISSUE #1181: [FEATURE] Hide H4ck, Dagna and Uma
## 🚀 Feature
Since the goal is to only have characters with the updated model (currently, Muflus is the only one), we need to remove H4ck, Dagna, and Muflus from the character selection, info, and main screen models

BertovDev:
Closes #1181 

## Motivation

Simplify the way the display the usable characters in the client ( MainScreen models, CharacterList and characterInfo)

Now `MainScreenManager` contains a list of strings `enableCharactersName`, which contains the names of the characters you want to be avaible. This is the variable you should modify to remove or add avaible characters.
In the next iteration we should map this with the json from the backend.

Captura de pantalla 2023-11-27 a la(s) 12 20 55

## Summary of changes

Backend:

1. Left only muflus as a playable character

Client:

1. Changed `MainScreenManager`, `UIModelManager`, `CharacterInfoManager` and `CharacterListManager`.
2. Add the method to `Utils` class.
3. Rename all Character UI models variants from name + Variant (ex: Muflus Variant) to just name (ex: Muflus), i made this to simplify the matching from models names and real character names
4. Rename H4ck model name from `H4ck_v2` to `H4ck`

## How has this been tested?

1. In the MainScreenManager script Add/Remove character names in the `enableCharactersName` list
2. Start the game, check if the characters the Add/Remove are avaible or not in the MainScreen models
5. Check if in the CharacterList scene you can click the characters that are in the list and check if you cannot click the characters that are not part of the list
6. Check if in the CharacterInfo screen you  only see the avaibles characters, use the arrows to see them all
7. Finally start a game and you only have to be able to play with muflus, since is the only avaible in the backend

## Test additions / changes

List tests added/updated.

## Checklist
- [x] This change requires new documentation.
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -8,13 +8,25 @@ public class UIModelManager : MonoBehaviour
...
     List<GameObject> playerModels;
 
+    [Tooltip("Enable characters name to be used")]
+    private List<string> enableCharacters = new List<string>();
+
     public void SetModel(CoMCharacter character = null)
     {
-        int index = Random.Range(0, playerModels.Count);
-        string name = playerModels[index].name;
+        List<string> avaibleModels = new List<string>();
```
ncontinanza, line 20:
```suggestion
        List<string> availableModels = new List<string>();
```

server/lib/dark_worlds_server/engine/engine_runner.ex
ncontinanza, line 1: You should move this changes to the runner.ex in the game_backend project

ncontinanza (changes requested):

tkz00 (changes requested):
Nice! Looks good, I think "enableCharacters" & "enableCharactersName" are not good names for fields, I personally think "available" is a better option than "enable", but even in that case, it shouldn't be "enable" it should be "enabled".

tkz00 (approved):
LGTM

BertovDev:
> Nice! Looks good, I think "enableCharacters" & "enableCharactersName" are not good names for fields, I personally think "available" is a better option than "enable", but even in that case, it shouldn't be "enable" it should be "enabled".

Nice! changed to `enabledCharactersName`

ncontinanza (approved):

Issue #905, opened 2023-09-18
[FEATURE] User Accounts


marco-paulucci:
### Authentication Methods
- DeviceId (unsecured)
- Email & Password
- Google
- Apple Game Center
- Facebook

### Login

The login flow should work as follows:

- 1. Check for any locally saved secure login methods and log-in automatically
- 2. If none is found, check for unsecured accounts with `deviceId`
- 3. If none is found, register a new account and login automatically

Additionally, the user should have a login option in their account settings to swap accounts using one of the Authentication services.

### Registration

The first time a player opens the game, a new account using their `deviceId` should be created. This means that players are ready to play as soon as they load the game, without friction. These accounts should be considered unsecured until the user adds any other authentication method(s). Users will be unable to do any IAP with unsecured accounts.

In a new user account page, the user will then be able to secure their account by linking it to one of the Authentication services mentioned above. Securing your account should clear any linked `deviceId` from the account, preventing logging into it just by owning that device. 

___

### Tasks

- [ ] Implement a UserAccounts entity and database
- [ ] Implement Registration methods as outlined above, for:
  - [ ] DeviceId
  - [ ] Mail/Password (+password recovery)
  - [ ] #60 
  - [ ] Apple Game Center
  - [ ] Facebook
- [ ] Implement Login methods for the selected Authentication services
- [ ] Implement DisplayNames
- [ ] Implement an "Accounts" menu in the frontend with all required functionalities

PR #1237, opened 2023-12-01
[GH-905] User creation on title screen

tkz00:
Works towards completing #905 

## Motivation

We need the player to have a registered user to store information about they.

## Summary of changes

In the loading screen after the title screen scene it's verified if there is a user in the backend associated with the device_id of the client, if there is not a user will be created.

Also in that check modifications have been added to check if Brazil server is responding (or the server which is default in that moment), and if it isn't, automatically start pointing towards localhost server. (If the loading spinner in title screen never disappears it is probable Brazil is unavailable and the localhost server isn't running)

## How has this been tested?

To test this PR correctly it is better to run the backend locally to have more visibility of changes and ability to edit.

To target the local server you need to edit the `_defaultServerIp` field of `SelectServerIP.cs` to "AAAA" (or anything random), so it doesn't target Brazil.

- Check a user is created in title screen: be sure to not have users with your device ID in your local database and go through until main screen scene, there if you check your database, you should see a new user with your device_id.
- If there already is a user in the database with your device_id, then no new user should be created.

## Test additions / changes

No tests have been added or updated.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,97 @@ public class TitleScreenController : MonoBehaviour
...
+        }
+    }
+
+    public void PlayButton()
+    {
+        ChangeToMainScreen();
+    }
+
+    private void ChangeToMainScreen() {
+        SetLoadingScreen(true);
```
agustinesco, line 70: I think the `PlayButton/0` method is not doing anything right?

agustinesco (changes requested):

agustinesco (dismissed):
LGTM! tested on localhost and the user creation is working as expected!

client/Assets/Scripts/Utils/Utils.cs
mFragaBA, line 1: I think it would be nice to have a BackendUtils.cs or similar so the requests to the backend logic doesn't get mixed up with the game logic itself

client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,93 @@ public class TitleScreenController : MonoBehaviour
...
+            error => {
+                switch(error) {
+                    case "USER_NOT_FOUND":
+                        CreateUser();
+                        break;
+                    case "CONNECTION_ERROR":
+                        // If server is not found default to localhost IP
+                        SelectServerIP.serverIp = "localhost";
+                        SelectServerIP.serverNameString = "Localhost";
+                        PlayButton();
```
mFragaBA, line 83: If the server's down, wouldn't this end up in an endless loop?

mFragaBA (changes requested):

client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,97 @@ public class TitleScreenController : MonoBehaviour
...
+                switch(error) {
+                    case "USER_NOT_FOUND":
+                        CreateUser();
+                        break;
+                    case "CONNECTION_ERROR":
+                        // If server is not found default to localhost IP
+                        SelectServerIP.serverIp = "localhost";
+                        SelectServerIP.serverNameString = "Localhost";
+                        ChangeToMainScreen();
+                        break;
```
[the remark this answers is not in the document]
tkz00: Added the message and fixed the infinite loop problem.

client/Assets/Scripts/Utils/Utils.cs
```diff
@@ -12,6 +15,14 @@ public class Utils
...
     public static readonly Color healthBarPoisoned = new Color32(66, 168, 0, 255);
 
+    class AcceptAllCertificates : CertificateHandler
+    {
+        protected override bool ValidateCertificate(byte[] certificateData)
+        {
+            return true;
+        }
+    }
+
```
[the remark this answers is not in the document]
tkz00: Removed the certificates assignation, continues to work well.

client/Assets/Scripts/Utils/Utils.cs
```diff
@@ -154,6 +165,114 @@ public static Gradient GetHealthBarGradient(Color color)
         };
     }
 
+    public static string MakeHTTPUrl(string path)
+    {
+        if (SelectServerIP.GetServerIp().Contains("localhost") ||
+            SelectServerIP.GetServerIp().Contains("10.150.20.186")
```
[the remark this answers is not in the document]
tkz00: Added the load tests servers.

client/Assets/Scripts/Utils/Utils.cs
tkz00: Moved the server connection logic to `ServerUtils.cs`. In the future it would be nice to have a static class with a generic request method that takes the route, http method and parameters, to avoid having a method for each request needed.

client/Assets/Scripts/Utils/Utils.cs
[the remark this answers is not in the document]
mFragaBA: Nice!

client/Assets/Scripts/Utils/Utils.cs
mFragaBA: Agreed!

client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,106 @@ public class TitleScreenController : MonoBehaviour
...
+    }
+
+    private IEnumerator LoadSceneAsyncProcess(string sceneName)
+    {
+        asyncOperation = SceneManager.LoadSceneAsync(sceneName);
+        asyncOperation.allowSceneActivation = false;
+        while (asyncOperation.isDone)
+        {
+            yield return null;
+        }
```
agustinesco, line 61: What's the purpose of this while cicle?

client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,106 @@ public class TitleScreenController : MonoBehaviour
...
+                },
+                error =>
+                {
+                    switch (error)
+                    {
+                        case "USER_ALREADY_TAKEN":
+                            Errors.Instance.HandleNetworkError("Error", "ClientId already taken");
+                            SetLoadingScreen(false);
+                            break;
+                        default:
```
agustinesco, line 121: Don't we also need a `playNowButton.EnableButton();` here?

Also if you are turning off the loading screen and enabling the play button on every switch case maybe you can put both actions outside of the switch

client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -22,21 +27,106 @@ public class TitleScreenController : MonoBehaviour
...
+                        case "USER_NOT_FOUND":
+                            CreateUser();
+                            break;
+                        case "CONNECTION_ERROR":
+                            Errors.Instance.HandleNetworkError(
+                                "Oops!",
+                                "No Server Avaible to Connect"
+                            );
+                            SetLoadingScreen(false);
+                            playNowButton.EnableButton();
```
agustinesco, line 89: same as above

client/Assets/Scripts/Utils/ServerUtils.cs
```diff
@@ -0,0 +1,127 @@
...
+public static class ServerUtils
+{
+    public static string MakeHTTPUrl(string path)
+    {
+        if (
+            SelectServerIP.GetServerIp().Contains("localhost")
+            || SelectServerIP.GetServerIp().Contains("10.150.20.186")
+            || SelectServerIP.GetServerIp().Contains("168.119.71.104")
+            || SelectServerIP.GetServerIp().Contains("176.9.26.172")
+        )
```
agustinesco, line 16: Can this if statment be simplified with an any statment? for what i can see the method `GetServerIp` returns an string, could we match that with any of the options? somethig like 
```
        List<String> servers = new List<String>{"localhost", "10.150.20.186", "168.119.71.104", "176.9.26.172"};
        if ( servers.Any((ip) => SelectServerIP.GetServerIp() == ip))
```

agustinesco (changes requested):

client/Assets/Scripts/UI/TitleScreenController.cs
BertovDev: You are right

client/Assets/Scripts/Utils/ServerUtils.cs
BertovDev: Sure!, this change improve the code readability but it make it more efficient? What do you think?

agustinesco: I think i'll be the same, we had 4 comparisons before the change , and now we have a list traverse of 4 elements with n comparisons

tvillegas98: We could improve efficiency by using a HashSet instead of a vector

agustinesco (dismissed):
LGTM!

client/Assets/Images/UI/loading-spinner.png
AngieDutra, line 1: This should be set to the size stablished by the team in Notion. For this sprite that is 512. I would also move these two sprites to a "Loading" folder

AngieDutra (changes requested):
When I finish a game and want to start another match the LobbyConnection is missing and i cannot do so

client/Assets/Images/UI/loading-spinner.png
BertovDev: Sure thing!

AngieDutra (approved):
Nice work!

AngieDutra (approved):
Really good job on this PR! LGTM ✨

agustinesco (approved):
Great! nice work @BertovDev  and nice catch there @AngieDutra

PR #1252, opened 2023-12-06
Fix spawn effect

BertovDev:
Closes #[issue]

## Motivation

Currently the spawn effect was desactivated. With this we activate and desactive it for all players in the init of the game.

## Summary of changes

Changed CharacterBase activateSpawnFeedback method to ToggleSpawnFeedback
Re-created the logic activation in CustomLevelManager `CameraCinematic` method

## How has this been tested?

Start a game you should see the spawnfeedback once at that time and not ever again.

## Test additions / changes

List tests added/updated.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/CharacterBase.cs
```diff
@@ -27,21 +27,19 @@ public class CharacterBase : MonoBehaviour
...
+        //     MMSoundManagerSoundPlayEvent.Trigger(
+        //         spawnSfx,
+        //         MMSoundManager.MMSoundManagerTracks.Sfx,
+        //         Utils.GetPlayer(SocketConnectionManager.Instance.playerId).transform.position
+        //     );
+        // } else {
+        //     sound3DManager.SetSfxSound(spawnSfx);
+        //     sound3DManager.PlaySfxSound();
+        // }
     }
```
agustinesco, line 44: Why are we keeping this code if we're not using it anymore?

client/Assets/Scripts/CustomLevelManager.cs
```diff
@@ -196,6 +201,19 @@ IEnumerator CameraCinematic()
             yield return new WaitForSeconds(2.1f);
             CancelInvoke("Substract");
             InvokeRepeating("MoveYCamera", 0.3f, 0.1f);
+            Utils
+                .GetAllCharacters()
+                .ForEach(character =>
+                {
+                    character.characterBase.ToggleSpawnFeedback(true, character.PlayerID);
```
agustinesco, line 208: Why not changing the `spawnFeedback` here?
```suggestion
                    character.characterBase.spawnFeedback.SetActive(true);
```

The same goes for the function call below

agustinesco (changes requested):
The spawn effect seems to be working as expected! nice.
I just have 1 question and some comments regarding the code.

With this change we are losing the spawn sounds, is this okey?

client/Assets/Scripts/CharacterBase.cs
```diff
@@ -22,24 +22,18 @@ public class CharacterBase : MonoBehaviour
     [SerializeField]
     Sound3DManager sound3DManager;
 
-    public IEnumerator activateSpawnFeedback(bool isCurrentPlayer)
+    public void ToggleSpawnFeedback(bool value, string id)
```
agustinesco, line 25: Can we rename the var value to something that indicate what we're receiving?? something like `activateSound` or `activateFeedback`

agustinesco (changes requested):
The sounds seems to be working as expected but whenever i enter a match i get this error
image

client/Assets/Scripts/CharacterBase.cs
BertovDev: NIce catch, i will change it.

BertovDev:
> The sounds seems to be working as expected but whenever i enter a match i get this error image

This message is totally expected. It happends beucause in this PR I changed the audio listener from the camera to the player, so until the characters are not created you will have this message send. Nothing to worry about and good that you notice it,

agustinesco:
> > The sounds seems to be working as expected but whenever i enter a match i get this error image
> 
> This message is totally expected. It happends beucause in this PR I changed the audio listener from the camera to the player, so until the characters are not created you will have this message send. Nothing to worry about and good that you notice it,

After talking about this we think that this is only a message and we should ignore it, you can read more about it [here](https://forum.unity.com/threads/there-are-no-audio-listeners-in-the-scene-thats-actually-fine-right.384596/)

agustinesco (dismissed):

client/Assets/Scripts/CharacterBase.cs
```diff
@@ -22,24 +22,18 @@ public class CharacterBase : MonoBehaviour
...
+        spawnFeedback.SetActive(isActiveSound);
+        if (isActiveSound)
         {
             MMSoundManagerSoundPlayEvent.Trigger(
                 spawnSfx,
                 MMSoundManager.MMSoundManagerTracks.Sfx,
-                Utils.GetPlayer(SocketConnectionManager.Instance.playerId).transform.position
+                Utils.GetPlayer(SocketConnectionManager.Instance.playerId).transform.position,
+                false,
+                0.01f
```
ncontinanza, line 35: We could have this value as a private constant, such as we did with the minimum volume in ToggleAudio.cs

ncontinanza (changes requested):

ncontinanza (commented):
Also, there's a cut in the game music during the "preparing battle" screen. The music stops, and after this screen it keeps on playing were it left

client/Assets/Scripts/CharacterBase.cs
BertovDev: Done!

ncontinanza (approved):
LGTM!

agustinesco (approved):
LGTM!

PR #1258, opened 2023-12-07
Items inventory

ncontinanza:
⚠️ This PR needs to be tested locally using the changes from https://github.com/lambdaclass/game_backend/pull/90

## Motivation

We need to be able to equip a picked up item, and use it afterwards.

## Summary of changes

When a player walks by an item, it's saved in the inventory and available to use.
When the inventory button is clicked, the item's effect is applied.

## How has this been tested?

Start a new game, and let your character receive some damage. Pick up a health item, and use it tapping on the inventory button. Notice how your health is now restored.

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


client/Assets/Scripts/Inventory.cs
```diff
@@ -0,0 +1,65 @@
...
+    [SerializeField]
+    public Sprite myrrasBlessing;
+    Image inventoryButtonImage;
+
+    private void Start() {
+        inventoryButtonImage = inventoryButton.GetComponent<Image>();
+    }
+    
+    private void Update() {
+        player = Utils.GetGamePlayer(SocketConnectionManager.Instance.playerId);
```
agustinesco, line 25: Why are we fetching the player every update? are the players being destroyed when they die?

agustinesco (dismissed):
LGTM! tested on local and there were no issues!

client/Assets/Scripts/Inventory.cs
ncontinanza: That's because we need to see the changes in the inventory, which is an attribute of the player. We're checking it every update to notice the change as soon as the inventory changes.

agustinesco (dismissed):
LGTM! tested on local and with an android build and it worked like a charm! great

client/Assets/Scripts/Inventory.cs
agustinesco: oooh right! it's a little bit controversial to traverse the player list on each frame but i don't see it as an issue, we sure will change it once we start to optimize the game. 

Thanks for explaining!

saugon:
@AngieDutra I'm having this error when using the item:
image

client/Assets/Scripts/Inventory.cs
```diff
@@ -0,0 +1,165 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+using DG.Tweening;
+using System.Collections;
+using Communication.Protobuf;
+
+public class Inventory : MonoBehaviour
+{
+    private const string MYRRAS_BLESSING = "loot_health";
+    private const string ITEM_FEEDBACK = "UseItemEffect";
```
saugon, line 11: Is there any reason these are hardcoded? Could they not be directly linked to the prefab instead?

AngieDutra: Agree, will do as requested!

AminArria (approved):
LGTM

agustinesco (approved):
LGTM! the items are looking amazing, great job guys. Tested both in the unity client and in an android build and worked as a charm

PR #1263, opened 2023-12-11
Character Info Updates

AngieDutra:
## Motivation

Updates were needed in the character info screen

## Summary of changes
- UI changes were made
- It was also added the victory animation to appear from time to time in character info and in main screen. Please check both
reference (4)

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -14,6 +14,8 @@ public class UIModelManager : MonoBehaviour
 
     [Tooltip("Enable characters name to be used")]
     private List<string> enabledCharacters = new List<string>();
+    bool animate = false;
+    float animationInterval = 20f;
```
BertovDev, line 18: `const float ANIMATION_INTERVAL = 20f;`

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -54,4 +58,16 @@ public void RemoveCurrentModel()
...
+
+    IEnumerator AnimateCharacter(GameObject modelClone)
+    {
+        while (animate)
+        {
+            yield return new WaitForSeconds(1f);
+            modelClone.GetComponentInChildren<Animator>().SetBool("CharacterInfo", true);
+            yield return new WaitForSeconds(1f);
+            modelClone.GetComponentInChildren<Animator>().SetBool("CharacterInfo", false);
+            yield return new WaitForSeconds(animationInterval);
```
BertovDev, line 70: This is well directed and its currently not working, I believe the following approach could be good to improve this:

1. We use the Victory animation that we already have, we dont need a new one.
2. Remane the clip to "Victory" so later when we have more characters we can use this for every one of them.
3.
```
  yield return new WaitForSeconds(1f);
            modelClone.GetComponentInChildren<Animator>().SetBool("Victory", true);

            if (animationClipDuration == 0.0f)
            {
                animationClipDuration = modelClone
                    .GetComponentInChildren<Animator>()
                    .runtimeAnimatorController.animationClips.ToList()
                    .Find(clip => clip.name == "Victory")
                    .length;
            }

            yield return new WaitForSeconds(animationClipDuration);
            modelClone.GetComponentInChildren<Animator>().SetBool("Victory", false);
            yield return new WaitForSeconds(ANIMATION_INTERVAL);
```
With this implementation we wait one second then play the victory animation then wait the amout of time of the player (so it dont get cut in the middle) and finally we desactivate the animation and wait for the interval.

client/Assets/Animations/Muflus/Last/GR6_VictoryInfo.anim
BertovDev, line 1: There is any reason to create a new Animation and not reuse the one we already have?

BertovDev (changes requested):

client/Assets/Animations/Muflus/Last/GR6_VictoryInfo.anim
AngieDutra: The original animation is looped but your review gave me a workaround this issue too

client/Assets/Scripts/UI/UIModelManager.cs
AngieDutra: Done

client/Assets/Scripts/UI/UIModelManager.cs
AngieDutra: I changed this up a bit to move the assignment out of the loop. However, great input! Already implemented it ✨

BertovDev (approved):
So cleeeeean!

saugon (approved):

PR #1271, opened 2023-12-12
New Araban ring

gergahn:
## Summary of changes

this was before.

Screenshot 2023-11-30 at 19 30 55

This is after.
Screenshot 2023-12-11 at 18 18 36


BertovDev (changes requested):
Looks awesome! and remember to allways tick the static box if the GameObject is not doing any type of movement (static).

Also check if we want to objects to cast shadows.I believe in this case we do not want so you should change the on to off or if we want the shadow tick thr static shadow casting.

Captura de pantalla 2023-12-14 a la(s) 15 25 31
Captura de pantalla 2023-12-14 a la(s) 15 27 14

gergahn:
Thanks Bauti! I totally forgot about that. I will be more alert next time!

BertovDev (approved):
Awesome!

saugon (approved):
Nice!

PR #1272, opened 2023-12-12
Character list updates

### LINKED ISSUE #1209: [FEATURE] Character list improvements
## 🚀 Feature
1. Remove class filter.
2. Add 12 character slots using:
    - All the characters' portraits that are finished.
    - A generic portrait with the "?" sign.

AngieDutra:
Closes #1209 

## Motivation

Character List screen was outdated and needed a style update and a refactor functionality wise. 
The links depend on index position and not by name. This is not ideal behavior.

## Summary of changes
Style changes:

reference (5)

A code refactor was done too:
The general idea is that now we have a GameObject in MainScreen.unity called `CharactersScriptableObjects` were we reference the character's scriptable objects. This game object persists throughout the game and from here we should retrieve the characters information from now on. This makes the process of adding new characters easier since it only needs one reference for it to be applied to all the game screens.
- Available characters
  - Before: The available character, Muflus, was hardcoded in several scripts
  - Now: The characters name is stated in a private variable in `CharactersManager` to be replaced when the backend gives us that information
- Link system
  - Before: the link system was stablished by the character's list index in `CharacterList.unity` but there were multiple lists, both of character's scriptable objects and ui models. For example, the models shown were selected based on a prefab list and all the information from a scriptableObject list. This information was all associated by the index position. This was error prone because a change in the reference order could result in an error or mix up of information
  - Now: starting in MainScreen our "selectedCharacter" is shown (another PR to set this in a database is in progress). If we go to CharacterList and select a character it is based on the character's name assigned in a variable called `goToCharacterName`. The upgrade here is all the information required for each scene (prefabs, models, string data, etc) comes from the same list. If the reference order were to change this will only affect the character list display order but no bugs or mixups regarding the information shown in character info
- Left and Right arrows
  - Before: the left and right arrows were enabled but only having one character available, causing an animation bug due to refresh
  - Now: the left and right arrows will be enable if there are 2 or more available characters. The index value of a character is only used to show the previous or next character in the list
- Main screen link
  - Before: the character shown at the beginning was hardcoded
  - Now: the character shown is a variable string called `selectedCharacterName` to be replace when the database saves this data
  - Now: If you click on the model in main screen it will take you to it's character info. If you go back you should go to charactersList.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


ncontinanza (changes requested):
Nice job!
I just found a UI detail to fix:
image

ncontinanza (approved):

agustinesco (approved):
LGTM! tested both on local and in an android release and the character list looks amazing! great job @AngieDutra

PR #1274, opened 2023-12-14
Unlock leap skill

agustinesco:
Closes #[issue]

## Motivation

we've worked on the leap functionality and this pr will turn on the leap button on the client

## Summary of changes

Added cooldown for skill 1 container so the button is interactable again.
Also restore some initializer for the skill1

## How has this been tested?

start a match with this branch and the correct back end branch runing on local host and you should be able to jump and hit everyone around when you land

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


client/Assets/Scripts/Battle.cs
```diff
@@ -316,35 +318,36 @@ void UpdatePlayerActions()
         ulong skillDuration
     )
     {
+        Debug.Log("aber action" + playerAction);
```
ncontinanza, line 321:
```suggestion
```

agustinesco: whoops

ncontinanza (approved):

BertovDev (commented):
This is so good and i believe it will be better to bring more information about some parts of the PR like:

- Why SkillAnimationEvent was deleted ?
- What VfxStep actually does ?
- Why do you have changes in the Battle Scene?

But besides that the PR is solid, nice!

agustinesco:
@BertovDev 
- For what i know the SkillAnimationEvent was like a safenet to remove the animations if needed, now the skill.cs is the one doing it 
- It looks like the VfxStep is a struct to safe different parameters for an animation, and instantiate them in the skill_info module, i'm not entirely sure since this change was part of the @saugon rework of the feedback
- Whoops, the battle scene was changed adding the cooldown to the skill1 but that's already done in the CameraUI prefab, i'll restore it

ncontinanza (dismissed):

client/Assets/Scripts/ClientPrediction.cs
```diff
@@ -33,7 +33,8 @@ void removeServerAcknowledgedInputs(OldPlayer player, long timestamp)
     void simulatePlayerMovement(OldPlayer player)
     {
         // TODO check this
-        var characterSpeed = PlayerControls.getBackendCharacterSpeed(player.Id);
+        //var characterSpeed = PlayerControls.getBackendCharacterSpeed(player.Id);
```
agustinesco, line 36: i think we should remove this

client/Assets/Scripts/CustomCharacter.cs
```diff
@@ -8,6 +9,7 @@ public class CustomCharacter : Character
     [Header("Character Base")]
     [SerializeField]
     public CharacterBase characterBase;
+    public List<OldActionTracker> currentActions = new List<OldActionTracker>();
```
agustinesco, line 12: Since this list is used a los to check if the element is inside of it and we only have one ocurrence of each action maybe we could replace it with a hashset to optimize this section of the code, not blocking but it's a possible upgrade

saugon: Nice catch! Fixed 🚀

saugon (approved):
Great work! 👏🏻

BertovDev (approved):

PR #1362, opened 2024-01-26
Re-enable lobby counter

AminArria:
Closes #[issue]

## Motivation

We want to show live updates of the lobby getting filled

## Summary of changes

For now this only re-enables the fake counter we had

## How has this been tested?

Run the game

## Test additions / changes

List tests added/updated.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.

### DIFF
--- a/client/Assets/Scenes/TitleScreen.unity
+++ b/client/Assets/Scenes/TitleScreen.unity
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.3708708, g: 0.37838137, b: 0.35725543, a: 1}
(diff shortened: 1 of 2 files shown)


AngieDutra (approved):
LGTM ✨

BertovDev (changes requested):
I believe it would be better to add the `lobbyCapacity` by code instead of changing the gameObject in the scene because it can lead to unwanted changes. Look how as well as changing the `lobbyCapacity` also the `m_IndirectSpecularColor` was changed.

So we can undo the changes in the scene with
` git checkout main -- client/Assets/Scenes/TitleScreen.unity`
 and then in the `ServerConnection` script in the init method we can add 
        `this.lobbyCapacity = 10;`

What do you think ?

manucamejo (approved):

BertovDev (approved):

PR #1370, opened 2024-01-29
Re-enable zone shrinking

AminArria:
## Test additions / changes

Test with https://github.com/lambdaclass/mirra_backend/pull/204


BertovDev (changes requested):
This condition `if(Mathf.Abs(pos.x) < mapRadius && Mathf.Abs(pos.z) < mapRadius)` is never true since  `Mathf.Abs(pos.z) < mapRadius` is always false.

BertovDev:
> This condition `if(Mathf.Abs(pos.x) < mapRadius && Mathf.Abs(pos.z) < mapRadius)` is never true since `Mathf.Abs(pos.z) < mapRadius` is always false.

Fixed! We had to divide by 100 the radius send from the backend. Because there is a difference of x100 between the values.

BertovDev (approved):

PR #1383, opened 2024-02-01
Improve GameServerConnectionManager

BertovDev:
## Motivation

Improve `GameServerConnectionManager`

## Summary of changes

Move the  `IsGameCreated` co routine from Updare method to Start

The method includes a `yield return new WaitUntil(() => !String.IsNullOrEmpty(SessionParameters.GameId));` so it is not necessary to having it in the Update method. It will start wait for the condition and then `ConnectToSession` which calls `OnWebSocketMessage` and starts listen to messages.

Without this change we are calling the co-routine `IsGameCreated` each frame.

## How has this been tested?

To test it you can add a Log inside the `IsGameCreated` and `OnWebSocketMessage` Update message and see how many times it is called.
Then change the Coroutine back to the Update method and see the logs again.

## Test additions / changes

List tests added/updated.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.

### DIFF
--- a/client/Assets/Scripts/GameServerConnectionManager.cs
+++ b/client/Assets/Scripts/GameServerConnectionManager.cs
@@ -51,6 +51,7 @@ public class GameServerConnectionManager : MonoBehaviour
     void Start()
     {
         Init();
+        StartCoroutine(IsGameCreated());
     }
 
     public void Init()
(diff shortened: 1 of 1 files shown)


lotuuu (approved):
Works fine! Checked with your log method and `IsGameCreated` is only called once!

PR #1409, opened 2024-02-07
[VFX] VFX for H4ck skills

DmytroTkachukPingleStudio:
## Motivation
Add new projectile for Skill1 and new projectile for Skill2 with colliders so they call OnCollisionEnter event to ProcessCollision. Add effects for Dash skill along with logic to spawn separate effects into custom bone positions(like eyes or legs). Add new Attack indicators

## Summary of changes

- add new Skill1, Skill2 and Dash effects
- add logic for spawning separate effects for Dash effects
- add projectile collision logic
- add new Attack indicators

## How has this been tested?

Skill1, or the basic projectile, can be tested by casting this skill during the game. Skill2's effect needs to be set additionally because it still spawns a projectile for Skill1. Projectile collision can be tested by placing an object with a collider and the 'ProjectileObstacle' layer in the scene. The Dash effect needs to be set additionally because I haven't found any logic that spawns any effect on Dash. All attack indicators can be tested by casting skills that require indicators.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


AngieDutra (approved):
Looks good! ✨

PR #1440, opened 2024-02-19
Implement uma ultimate

agustinesco:
Now the player structure will include a map of effects 

When uma uses radiance she will have an effect with the name "invisible"


tkz00 (approved):
Works great!

I would further reduce the alpha value of the player's color, I tried a 0.3 and saw it better.

Also, in the future invisible players will not be sent in the gamestate each update to the clients that shouldn't be able to see them? I thought about this bc in the current state it's possible to write a script to see invisible players when you aren't supposed to, since you still get the invisible player's in the game state, just with an additional effect.

agustinesco:
> Works great!
> 
> I would further reduce the alpha value of the player's color, I tried a 0.3 and saw it better.
> 
> Also, in the future invisible players will not be sent in the gamestate each update to the clients that shouldn't be able to see them? I thought about this bc in the current state it's possible to write a script to see invisible players when you aren't supposed to, since you still get the invisible player's in the game state, just with an additional effect.

Yes! we've considered that this is easy to cheat but it won't be implement yet, the plan is to not send player that you shouldn't see

mgrunwaldt (approved):

BertovDev (approved):

PR #1473, opened 2024-02-21
Implement valtimer ultimate

agustinesco:
Test with this [Backend PR](https://github.com/lambdaclass/mirra_backend/pull/298)

Now when you pick valtimer and try to cast his ultimate you'll place an area that damage and pull enemies to his center

Todo list:
- [x] Implement a visual representation of the damage pools
- [ ] Adjust range and damage


BertovDev:
Tested on android s23 working fine.

BertovDev:
We should tweak the config to adjust the range and radius of the ability

Nico-Sanchez (changes requested):
It doesn't seem to work accurately. The ultimate's pool object doesn't spawn where I'm aiming and the visual representation size is not the same as the one received from the backend.
This can be seen by a character getting pulled in to a pool that it's not colliding yet in the visuals.

BertovDev:
> It doesn't seem to work accurately. The ultimate's pool object doesn't spawn where I'm aiming and the visual representation size is not the same as the one received from the backend. This can be seen by a character getting pulled in to a pool that it's not colliding yet in the visuals.

You are right! all adjust are done!

client/Assets/Scripts/Skills/Skill.cs
```diff
@@ -314,22 +315,26 @@ public float GetSkillRadius()
...
+                radius = skillInfo.skillCircleRange;
                 break;
         }
         return radius;
     }
 
     public void SetSkillRadius(float radius)
     {
-        skillInfo.skillCircleRadius = radius;
+        skillInfo.skillCircleRange = radius;
```
Nico-Sanchez, line 326: Shouldn't we rename this? Looks awkward to have a variable called "range" and assign a "radius" to it.

BertovDev: Yess, nice catch

BertovDev: Also renamed another vars.

BertovDev:
Draft till we decide what to do with the scaled textures inputs.

BertovDev:
> Draft till we decide what to do with the scaled textures inputs.

We will continue with this textures for now. Implementing a new one in another iteration.

Nico-Sanchez (dismissed):

AngieDutra (dismissed):
Tested on Desktop and Android and it's looking awesome!

AngieDutra (commented):
I'll say this prefab is changing on use and shouldn't
Captura de pantalla 2024-03-11 a las 16 48 20 p  m

AngieDutra (approved):
Tested with a match between two desktops and worked just fine ✨ No prefabs modified in the process

ncontinanza (approved):
LGTM!

Nico-Sanchez (approved):

PR #1489, opened 2024-02-23
Fix direction when moving

manucamejo:
Closes #[issue]

## Motivation

The player's direction was wrong because added a threshold to the magnitude of the movement between frames.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


agustinesco (changes requested):
The initial direction of the players is correct now but if i dash while holding the movement key the dash seems wrong

manucamejo:
> The initial direction of the players is correct now but if i dash while holding the movement key the dash seems wrong

Thanks, Agus, you're right.
I just fixed it.

agustinesco (approved):
LGTM!

BertovDev (approved):

PR #1492, opened 2024-02-23
Add zone shrink to update and fix zone startup

agustinesco:
Test this with [This Backend PR](https://github.com/lambdaclass/mirra_backend/pull/321)

We'll receive and display the time until the zone starts shrinking again


manucamejo (approved):

AngieDutra (changes requested):
The countdown works but it goes back to 20 once the zone is shown

Captura de pantalla 2024-02-26 a las 12 38 33 p  m

agustinesco:
> The countdown works but it goes back to 20 once the zone is shown

I didn't understood the comment, the zone timer works as this:
60 to 0 before the zone spawns, when the zone spawns the timer will stops at 20 and start going down to 0 when the zone starts shrinking again

AngieDutra (commented):
I'll take upon the implementation to fix this representation in the UI issue. Untill that is done this PR will be in draft

AngieDutra:
Talked with game design team and the timer backend logic was changed to fix the issue mentioned before. This resulted in no front end additions on my part so this PR is ready for review ✨

AngieDutra (approved):
Awesome! ✨

PR #1523, opened 2024-02-28
Fix client prediction starting position

manucamejo:
You will find the movement more smoothly (not robotic for the others and without the threshold movement for you).
Also, this PR fix the bug at the beginning of the game.


AngieDutra (commented):
✔️ Tested on Android Samsung S23+
✔️ Starting position is good
- Checked the movement in my computer and mobile but could see a significant difference. Is there a way to fact check this improvement?

BertovDev (approved):

AngieDutra (commented):
I'll merge main here since I see some build setting errors

AngieDutra (approved):
Awesome!

PR #1537, opened 2024-03-01
PostProcessing not affecting UI in game

### LINKED ISSUE #1532: [BUG] Post processing is affecting UI indicators in game
## 🐛 Bug Report

The lightning in game changes the ui element colors. This should not be the case

### Steps to Reproduce
Toggle the post processing in the battle scene

https://github.com/lambdaclass/curse_of_mirra/assets/82987608/33d31aca-4e6a-485a-b437-bf7f1da20519

### Posible solution

Maybe if the UI layer is set to be ignored by the post process effect and these elements are included in that layer this issue would be solved

AngieDutra:
Closes #1532 

## Motivation

The UI in game was changing their color because it was affected by post processing

## Summary of changes

The camera layers were wrong so that was fixed

A new layer called "NoPostProcess" was created and applied to all desired elements we don't want to be affected by the post processing component in game

Check:
- stamina indicators
- orientation indication of player
- player name
- health bar
- skill indicators

The reason for not applying that to the feedbacks is I believe the postprocessing there makes the feedback feel more as if they were part of the world and character instead of being clearly in another layer.

## How has this been tested?

In main you can see this:

https://github.com/lambdaclass/curse_of_mirra/assets/82987608/93a42be3-b212-4f59-a808-08dd8206c35a

Take into account you can't see the fix in the scene but it does apply in the simulation and mobile build:

[ Top part > Scene // Bottom part > Simulation ]
Captura de pantalla 2024-03-01 a la(s) 13 26 56

[Mobile build]
Screenshot_20240301_135930_client

## Test additions / changes

List tests added/updated.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.

### DIFF
--- a/client/Assets/Prefabs/AimDirection.prefab
+++ b/client/Assets/Prefabs/AimDirection.prefab
@@ -10,7 +10,7 @@ GameObject:
   m_Component:
   - component: {fileID: 6787156628499368461}
   - component: {fileID: 6738943573842909003}
-  m_Layer: 10
+  m_Layer: 24
   m_Name: AimDirection
   m_TagString: Untagged
   m_Icon: {fileID: 0}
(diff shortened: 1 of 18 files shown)


BertovDev (approved):
Amazing

PR #1543, opened 2024-03-05
UI buttons cancel action

### LINKED ISSUE #1289: [REFACTOR] Cancel action if pointer is released outside button
## 🚧 Refactor

Currently we have `ButtonAnimationsMMTouchButton.cs` controlling buttons animations and actions. 
If the user pressed the button and then it drags the finger outside the button and releases it, the action is still trigger. UX conventions indicate this would cancel the button's actions. This is the issue to resolve.

### Implementation

This implementation was already started in PR [1209](https://github.com/lambdaclass/curse_of_mirra/pull/1272) in [ButtonAnimationsMMTouchButton.cs](https://github.com/lambdaclass/curse_of_mirra/pull/1272/files#diff-421929f3a7d4d36946b6f371cee5f786dcb1bf8c5c72c15a42b00d85d7d4dc2a) and a couple of buttons such as the character model link in main screen ([MainScreenManager.cs](https://github.com/lambdaclass/curse_of_mirra/pull/1272/files#diff-0c0b69d093b73f9484a373675dc996574c2b93b364228b39ada8c242c91b293a)). However, this should be implemented all across the ui buttons so check those files out and you can get started!

Related: https://github.com/lambdaclass/curse_of_mirra/issues/1194 and https://github.com/lambdaclass/curse_of_mirra/issues/1147

### LINKED ISSUE #1542: [REFACTOR] Cancel UI buttons actions taking the container size as reference
## 🚧 Refactor

If the user moves a bit the finger over a container but the release is still over it id does not work the action trigger

### Implementation

The action shouldn't trigger if the finger is released outside the button but it should work if it's inside, even though the finger is moved a little

AngieDutra:
Closes #1289 
Closes #1542 

## Motivation

The intuitive ux was lacking a cancel action when releasing a press pointer trigger outside the button

## Summary of changes

Two logics are implemented:
- [Character list] List elements use a release difference to check if the element was selected or if the press detectedwas part of a scroll motion
- [All UI buttons with action release] Checks if the pointer is inside or outside the button to display the feedback animations and the actions accordingly.

## How has this been tested?

Check all UI buttons are working as expected. Try to test this also from the point of view of a user that starts pressing on an option they want to cancel the trigger of. It should be intuitive dragging the finger out of said button and have nothing happen.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/UI/ButtonAnimationsMMTouchButton.cs
```diff
@@ -67,27 +81,40 @@ public override void OnPointerUp(PointerEventData eventData)
...
+        else if (isInsideCard && !listElement)
         {
-            executeRelease = true;
+            return executeRelease = true;
         }
         else
         {
-            executeRelease = false;
+            return executeRelease = false;
         }
```
BertovDev, line 111:
```
      if (isInsideCard && listElement)
        {
            executeRelease = touchXDifference < MIN_DIFFERENCE && touchYDifference < MIN_DIFFERENCE;
        }
        else
        {
            executeRelease = isInsideCard && !listElement;
        }
        return executeRelease;
```

client/Assets/Scripts/UI/GoToCharacterInfo.cs
```diff
@@ -0,0 +1,13 @@
...
+public class GoToCharacterInfo : MonoBehaviour
+{
+    public string characterNameString;
+
+    public void GoToCharacterSelection()
+    {
+        CharactersManager.Instance.SetGoToCharacter(characterNameString);
+        GetComponent<MMLoadScene>().LoadScene();
+    }
+}
```
BertovDev, line 13: I believe we can avoid to create this class only for this functionality, instead we can create a method in Utils. I have two options let me know which one fits better for you.

Option 1
In `Utils.cs`
```
  public static void GoToCharacterInfo(string characterName){
        CharactersManager.Instance.SetGoToCharacter(characterName);
    }
```
in `MainScreenManager.cs`

```
   public void GoToCharacteInfo(){
        Utils.GoToCharacterInfo(characterSelectedName);
        GetComponent<MMLoadScene>().LoadScene();
    }
```
An call this method in the button

Option 2
In `Utils.cs`
```
  public static void GoToCharacterInfo(string characterName, string sceneName){
        CharactersManager.Instance.SetGoToCharacter(characterName);
        SceneManager.LoadScene(sceneName);
    }
```
in `MainScreenManager.cs && CharacterListItem.cs`

```
   public void GoToCharacteInfo(){
        Utils.GoToCharacterInfo(characterSelectedName, sceneNameStringHere);
    }
```
With the option 2 we could remove the `MMLoadScene` component in the affected GameObjects.

BertovDev (changes requested):
Really good, only leaving some comments.

client/Assets/Scripts/UI/GoToCharacterInfo.cs
AngieDutra: I like the second approach!

BertovDev (approved):

agustinesco (approved):
LGTM!

PR #1584, opened 2024-03-13
Minor QoL improvements

AminArria:
This PR introduces 2 changes that need to be tested in builds which means I can't test them and need help :D 

- Version number, text showing version should be something like `20240313201905-pre-alpha` which is roughly `year month day hour minute second`. We could simplify a bit of the text, so feel free to comment on it. Also, we could expand the space for the text
- Default server, when building in Unity editor it should default to localhost, but when making builds they should default to Brazil server (was the current setup)


BertovDev (changes requested):
BuildTimestamp not working on mobile Build.

BertovDev:
Implemented server selection QoL in #1611

PR #1590, opened 2024-03-14
Update messages.proto for player cooldowns

AminArria:
This PR updates `messages.proto` to have the new player cooldowns sent from the backend, but it does nothing with that information


BertovDev:
I did this in #1595. Sorry I did not see it before.

PR #1591, opened 2024-03-14
Fix character rotation

manucamejo:
## Motivation

We felt like the player slides when changing direction.

## Summary of changes

We were using the joystick value to calculate the character rotation in every frame.
On the other hand, we only accept movement every 30ms.
So the rotation was out of phase with the movement.

We did 2 changes here:
- Add a forced movement message to the backend every 500ms.
- Use the player direction to calculate the character rotation instead of the joystick value. 

## How has this been tested?

Playing with muflus in several servers.

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.

### DIFF
--- a/client/Assets/Scripts/Battle.cs
+++ b/client/Assets/Scripts/Battle.cs
@@ -32,6 +32,7 @@ public class Battle : MonoBehaviour
     private bool playerMaterialColorChanged;
     private bool sendMovementStarted = false;
     private long lastMovementUpdate;
+    private long lastForcedMovementUpdate;
 
     [SerializeField]
     private CustomLevelManager levelManager;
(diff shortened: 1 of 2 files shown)


BertovDev (commented):
I do not find the movement smooth.

https://github.com/lambdaclass/curse_of_mirra/assets/62818001/9f6fd988-3701-418d-b1d1-79daa7a6096c

What do you think?

Nico-Sanchez (approved):
After several tests and comparing this branch with main. I see that this PRs doesn't introduce a new bug and it solves the problem where the character moves to a certain direction while looking to a different one.

Nico-Sanchez:
> I do not find the movement smooth.
> 
>  Screen.Recording.2024-03-15.at.09.44.58.mov 
> What do you think?

Do you have a video from main to compair?

saugon (approved):

PR #1592, opened 2024-03-14
New healing vfx

### LINKED ISSUE #1011: [FEATURE] Healing vfx
Ditto

saugon:
Closes #1011 

## Motivation
The previous vfx didn't feel appropriate for the healing effect.

## Summary of changes
- Created a new particle effect with its own material and texture.
- Created a prefab and added it to the FeedbackContainer prefab.
- Linked the feedback prefab to each character.
- Renamed the previous vfx to AbsorbFeedback (since it gives that sensation and could be useful to keep).

## How has this been tested?
Not performing any skill after receiving damage will trigger the auto-healing process and therefore the vfx.

## Reference
https://github.com/lambdaclass/curse_of_mirra/assets/6811225/38876677-c131-4272-bd2d-4ed3b6ab331f

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


AngieDutra (changes requested):
Tested with all characters. 
The vfx seems a bit out of system but it's a clear visual representation of what happens. 
I would suggest moving the game object a bit up. I would also reduce a bit the max size of the crosses
Captura de pantalla 2024-03-14 a la(s) 8 03 04 p  m

In Muflus case it seems as the effect is not fully getting out of his body and it looks not as clean as it could


https://github.com/lambdaclass/curse_of_mirra/assets/82987608/a8e61219-ea4f-4bd5-8963-8bb04650b2e6



Overall, maybe moving the particles a bit up in the Y axis and reducing just a bit the crosses's max sizes it could look more smooth. Also, check the effect duration because imo if the player's health is very low and the healing process starts, once the character's health bar is full the animation is still going for a little bit longer that it should

saugon:
I made a few tweaks based on your comments @AngieDutra 👍🏻

saugon:
> I still think maybe the position of the game object + the radius and scale of it could improve the spawn effect of the crosses to be more outside of each character. In uma's case is working but with muflus it kind of spawns not as outside the model as it should
> Healing.feedback.mov
> 
> Also, I believe the use item feedback got broken in this PR
> Screen_Recording_20240321_114538_client.mp4

I agree with the proposal. I decided to go the other way because maybe it had too much effect, and we could iterate if it doesn't work, but let's start by trying the adjustments you mentioned 🚀 

About the item feedback, i'm not having that issue. Could you try restarting Unity and see what happens?

AngieDutra (commented):
I tested, closed unity and re tested but i'm still not seeing the vfx for using an item

saugon:
> I tested, closed unity and re tested but i'm still not seeing the vfx for using an item Captura de pantalla 2024-03-21 a la(s) 4 22 13 p  m

Do you have all the ThirdParty dependencies?
image

AngieDutra:
I'll re-import the assets in case that's the issue. I believe so

AngieDutra (approved):
The asset was broken only locally! My bad, awesome PR ✨

BertovDev (approved):
LGTM

PR #1607, opened 2024-03-19
Fix countdown time

### LINKED ISSUE #1601: [BUG] The client is not following the countdown properly
https://github.com/lambdaclass/curse_of_mirra/assets/20927859/0387201d-0e7f-4a8a-8f79-e24a94250721

BertovDev:
Closes #1601 

## Motivation

The prepare for battle animation was still playing even when the gameStart countdown from backend already finished. 

## Summary of changes

Adjust times, and coroutines exit statement. 

## How has this been tested?

Start different games, corroborate all of them starts in the correct time. You can check this with the danzer zone timer in the matchStats container. It should appear as 35 seconds (That is the value it has in the config)

## Test additions / changes

List tests added/updated.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.

### DIFF
--- a/client/Assets/Scripts/UI/PrepareForBattleAnimations.cs
+++ b/client/Assets/Scripts/UI/PrepareForBattleAnimations.cs
@@ -35,10 +35,12 @@ public class PrepareForBattleAnimations : MonoBehaviour
     Vector3 playerPosition;
     GameObject player;
     const float CAMERA_START_OFFSET = 30f;
-    const float PREPARE_FOR_BATTLE_DURATION = 4f;
-    const float CHARACTERS_DISPLAY_DURATION = 5f;
+    const float PREPARE_FOR_BATTLE_DURATION = 3f;
+    const float CHARACTERS_DISPLAY_DURATION = 4f;
     float TIME_UNTIL_GAME_STARTS = 0f;
     const float SURVIVE_DURATION = 1.65f;
+
+    bool countdownDone = false;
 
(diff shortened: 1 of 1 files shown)


agustinesco (changes requested):
This solution is not working as expected, if i start a match with my phone and the unity client i'm having a difference of ~3 seconds almost every time

BertovDev:
> This solution is not working as expected, if i start a match with my phone and the unity client i'm having a difference of ~3 seconds almost every time

I believe this is due to the graphic power of the devices. Since we are not having in count the graphic power of each phone and we do not know when each player phone already charged everything

BertovDev:
> > This solution is not working as expected, if i start a match with my phone and the unity client i'm having a difference of ~3 seconds almost every time
> 
> I believe this is due to the graphic power of the devices. Since we are not having in count the graphic power of each phone and we do not know when each player phone already charged everything

Maybe we should wait until each player is done charging the graphic things.

AngieDutra:
I tested this and worked fine in Android build. I'm waiting for the @agustinesco to check his change request until I approve this

agustinesco (approved):
LGTM! when my phone load fast enough the timers are synced, maybe we could work in a future for a fix for lower end phones

Tested on an android build and the unity client and created an issue to fix the timers for the lower end phones #1616

AngieDutra (approved):
Awesome! Tested on android build too ✨

PR #1619, opened 2024-03-21
[GH-1598] Teleport when distance between client and server positions are too high

### LINKED ISSUE #1598: [FEATURE] Teleport character when server doesn't answer in time
## 🚀 Feature

Teleport characters when server takes time to answer messages (high ping, etc.).

### Motivation

At this moment, when the server doesn't send any message and then we get one, the characters moves smoothly from the current position to the one the server just sent us. This takes time to let the user know their current position. A better solution to this could be teleporting them if a server msg takes a considerable amount of time.

### Additional context
We could add a video here.

Nico-Sanchez:
Closes #1598 

## Motivation

Conciliating the backend and frontend positions takes too much time if we make the model walk from one to another.

## Summary of changes

- When distance is higher than the threshold, teleport the player.
- Do not teleport other models that are not the player (client prediction and interpolation ghosts).

## How has this been tested?

Toxiproxy doesn't work for this. It delays the server answer but it keeps receiving the client messages normally.
The real case is when your network is not working properly, you can connect to a wifi that is too far or a poor network if you can.
We found a better workaround to test this (thank toxiproxy's): whenever you send a move message to the backend, we "sleep" the backend process so it doesn't answer in time and doesn't take any messages as well. You can achieve that using the backend branch -> [test-teleport-on-high-latency](https://github.com/lambdaclass/mirra_backend/tree/test-teleport-on-high-latency).

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


BertovDev:
Reviewing

BertovDev:
Seems to be working as expected. Tested in different environments: 

- Europe with 240ms : Seems to be working good. 240ms a playable amount of lag so no teleport was been processed
-  With the backend this PR suggest.  Seems to work as expected , teleporting the player when the difference between the client is too much.

Maybe when the latency is too high we should freeze the player, show a re-connection popup and then teleport the player to the real position.

BertovDev (approved):
Working fine to me. Things to have in count to next development:

- When the connection is too bad and the difference between client inputs and server is too far we should:

1. Freeze the client 
2. Show  recconection popup or something to refer to it
3. Wait until the server reconciliate its position
4. Un-freeze the client when both positions (server and client) are the same.

PR #1626, opened 2024-03-22
faster basic anims

marco-paulucci:
all basics animations are faster to make the game more responsive. 
relies on a [backend PR ](https://github.com/lambdaclass/mirra_backend/pull/398)too so this won't work on its own.

### DIFF
--- a/client/Assets/Animations/Uma/UmaController.controller
+++ b/client/Assets/Animations/Uma/UmaController.controller
@@ -672,7 +672,7 @@ AnimatorState:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_Name: Tr_Avenge
-  m_Speed: 1.5
+  m_Speed: 1.35
   m_CycleOffset: 0
   m_Transitions:
   - {fileID: -5522672172395918497}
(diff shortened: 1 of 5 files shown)


client/Assets/Animations/Uma/UmaController.controller
```diff
@@ -672,7 +672,7 @@ AnimatorState:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_Name: Tr_Avenge
-  m_Speed: 1
+  m_Speed: 1.5
```
BertovDev, line 675: Seems to much for me, maybe 1.3 would be better. What do you think?

BertovDev (commented):
Feels nice and only Uma basic seems to be too much in my opinion.

What do you think?

AngieDutra:
Read the description. This PR no longer depends on backend since this back [PR](https://github.com/lambdaclass/mirra_backend/pull/426) is merged

AngieDutra (approved):
Tested in Unity desktop and Android build ✨

BertovDev (approved):
Nice! once Valtimer is totally added we should do the same for him.

PR #1637, opened 2024-03-25
Add bush mechanic

agustinesco:
Test with [This Backend Branch](https://github.com/lambdaclass/mirra_backend/pull/417)


AngieDutra:
This PR is in draft because of changes in priorities. It may be best to just close and re-do this one

Issue #1645, opened 2024-03-30
[BUG] Responsive issues in death splash


AngieDutra:
## 🐛 Bug Report

The Models in UI in death splash seem to be having responsive issues

### Steps to Reproduce

This was found in an Android Galaxy A-03 but the same error can be replicated using in the simulator the following device:

<img width="162" alt="Captura de pantalla 2024-04-02 a la(s) 8 18 15 p  m" src="https://github.com/lambdaclass/curse_of_mirra/assets/82987608/3ee99bb3-e3b6-43db-9750-f16a588d2d43">

#### Screenshot

![WhatsApp Image 2024-03-30 at 13 27 37 (1)](https://github.com/lambdaclass/curse_of_mirra/assets/82987608/ac28e356-d225-406e-b2ea-a0a625ad15d1)
![WhatsApp Image 2024-03-30 at 13 27 37](https://github.com/lambdaclass/curse_of_mirra/assets/82987608/1c2b0694-3756-439b-9969-1201fa0e76c9)

PR #1646, opened 2024-04-03
Fix death splash sprites

### LINKED ISSUE #1609: [BUG] Fix rank assets
## 🐛 Bug Report

The progress container for the rank and its progress bar are stretched to the right.

### It's like this: (reference)
IMG_3767

IMG_3841

### It should look like this: (reference)

reference-win

### Assets 
https://drive.google.com/drive/folders/1sXvQomDqcO2NwPdhtzq6U_7qglZkGmWP?usp=drive_link

AngieDutra:
Closes #1609 

## Motivation

The death splash sprites were distorted, there were some issues in the references itself also. Now that erverything was sorted out the implementation could be smoother

## Summary of changes

This PR fixes:
- Those sprite distortion issues
- The character UI shadow now is sized by the character model both in character info screen and death splash
  - keep in mind the animations need a refactor. The shadows will only look good for the standing poses but will look weird when the characters falls down (like in muflus's and uma's defeat animation) this is a known issue that needs to be addressed by the design and animation team. Therefore, that is out of the scope of this PR
  
Screenshots:

Captura de pantalla 2024-04-03 a la(s) 12 39 53 a  m
Captura de pantalla 2024-04-03 a la(s) 10 00 07 a  m

## How has this been tested?

Go to the different screens available in the game were the character model is shown in the UI and test the new additions. Try winning and loosing to check everything is as expected

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -19,15 +24,38 @@ public void SetModel(string characterName)
...
+        }
+        if (characterShadow != null)
+        {
+            SetCharacterShadow(characterName);
+        }
+    }
+
+    public void SetCharacterShadow(string characterName)
+    {
+        if (characterName == "H4ck")
```
agustinesco, line 43: i'm not a huge fan of this approach, i think this will changed in the future when we implement a proper implementation for shadows (this is something mentioned in the pr description)

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -61,4 +89,42 @@ IEnumerator AnimateCharacter(GameObject modelClone)
...
+    {
+        if (Utils.GetCharacter(GameServerConnectionManager.Instance.playerId))
+        {
+            bool isWinner = GameServerConnectionManager
+                .Instance
+                .PlayerIsWinner(GameServerConnectionManager.Instance.playerId);
+            string animationName = isWinner ? "Victory" : "Defeat";
+            if (modelAnimator.parameterCount > 0)
+            {
+                bool hasAnimationParameter = AnimationHasParameter(animationName);
```
agustinesco, line 103: The method `AnimationHasParameter/1` is used only here and the passed to the method `HandleAnimation/2`, i think we should delete this method and include the check for the existence of the animation in the `HandleAnimation` method

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -61,4 +89,42 @@ IEnumerator AnimateCharacter(GameObject modelClone)
             yield return new WaitForSeconds(ANIMATION_INTERVAL);
         }
     }
+
+    public void ShowCharacterAnimation()
```
agustinesco, line 93: I think we should rename this since we're not showing any animation, we are showing the end of game animation, maybe `ShowEndGameCharacterAnimation` is more suitable

client/Assets/Images/UI/BattleUI/EndGame/reference defeated new.jpg
agustinesco, line 1: I'm pretty sure this is not supposed to be here right?

agustinesco:
The shadows seems to be working as expected, nice! i have one question tho, why are we using different shadows for the character info and the end game screen? 

End game: 
image

Character info:
image

The character info one has some kind of blur at the end and the one from the end screen doesn't

client/Assets/Scripts/UI/UIModelManager.cs
BertovDev: Yep, i like this approach

client/Assets/Scripts/UI/UIModelManager.cs
BertovDev: Yes this is correct, since we already have another method to play the animations in the characterInfo (`AnimateCharacter/1`). Also makes me think of a possible refactor to unify them.

client/Assets/Images/UI/BattleUI/EndGame/reference defeated new.jpg
BertovDev: Nope, deleting it now.

client/Assets/Scripts/UI/UIModelManager.cs
BertovDev: Yep, not a big fan also. I changes it adding a scale values of each character shadow in the character ScriptableObject and then assign it by script. 

Let me know if you see it better.

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -116,8 +93,9 @@ private bool AnimationHasParameter(string parameterName)
         return param != null;
     }
 
-    public void HandleAnimation(string animationName, bool hasAnimationParameter)
+    public void HandleAnimation(string animationName)
     {
+        bool hasAnimationParameter = AnimationHasParameter(animationName);
```
agustinesco, line 98: what i meant was to delete the `AnimationHasParamerter` method, the end result would be something like:
```suggestion
        bool hasAnimationParameter = modelAnimator
            .parameters
            .ToList()
            .Any(p => p.name == parameterName);
```

agustinesco:
I think the scale of the shadow is not quite right now

image

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -61,4 +67,42 @@ IEnumerator AnimateCharacter(GameObject modelClone)
...
+    private bool AnimationHasParameter(string parameterName)
+    {
+        AnimatorControllerParameter param = modelAnimator
+            .parameters
+            .ToList()
+            .Find(p => p.name == parameterName);
+
+        return param != null;
+    }
+
```
agustinesco, line 95: Obliterate this

```suggestion
```

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -61,4 +67,32 @@ IEnumerator AnimateCharacter(GameObject modelClone)
...
+            if (modelAnimator.parameterCount > 0)
+            {
+                HandleAnimation(animationName);
+            }
+        }
+    }
+
+    public void HandleAnimation(string animationName)
+    {
+        bool hasAnimationParameter = modelAnimator.parameters.ToList().Find(anim => anim.name == animationName) != null;
```
agustinesco, line 88: i think using `Any` is cleaner than comparing with null right?

```suggestion
        bool hasAnimationParameter = modelAnimator.parameters.ToList().Any(anim => anim.name == animationName);
```

client/Assets/Scripts/UI/UIModelManager.cs
agustinesco: hmmm yeah i think that may work

agustinesco (approved):
LGTM! tested on the unity editor and in an android build

AngieDutra:
Awesome, will review now ✨

AngieDutra:
All done! The model and shadow were in a weird position and scale. I adjusted that and checked Valtimer since he is now in main

agustinesco (changes requested):
image

i think the models are flying in the end screen, tested in the simulator with the google pixel 5

AngieDutra:
There is a responsive error reported in issue #1645 but I think I know how to fix this. Will try to do it here to avoid bugs in main

AngieDutra:
I did a small fix also solving issue #1645 but this should be refactor. Therefore, I'll leave the issue open to address this later on. For now I believe this approach works until the refactor takes priority

client/Assets/Scripts/UI/UIModelManager.cs
```diff
@@ -17,6 +18,14 @@ public class UIModelManager : MonoBehaviour
     Coroutine characterAnimation;
     Animator modelAnimator;
 
+    void Start()
+    {
+        if (rawImage)
+        {
+            rawImage.transform.localScale = ResponsiveModel();
+        }
```
BertovDev, line 26: I do not like this approach of let the others SerializeFields in null it seems like somethings is missing. Instead if you want these change only for the death Splash you can do the same thing but in `EndGameManager`.

What do you think?

BertovDev (changes requested):
Responsiveness seems to be working well! just a small comment.

client/Assets/Scripts/UI/UIModelManager.cs
AngieDutra: I dont think that's the solution to the issue you mention. We could add a bool to have the input field be hidden or visible but I believe it should be in the model manager in the final refactor. wdyt about this approach?

BertovDev: Yep this could be another solution. Nice

agustinesco (dismissed):
LGTM! the model seems to be correct in the two mobile version that you commented

BertovDev (approved):
Nice!! 

Death splash will be fixed in https://github.com/lambdaclass/curse_of_mirra/issues/1645

AngieDutra:
Let me check something with our TL before merging this

AngieDutra:
The character shadow in the death splash gives away our responsive position issue. The shadow was taken out of this PR and we will look into this later on. I did this [issue](https://github.com/lambdaclass/curse_of_mirra/issues/1684) to keep track

BertovDev (approved):
LGTM

alfopisano (approved):
Awesome!! Tested on iOS build and Unity desktop

PR #1652, opened 2024-04-05
[VFX] Added Magic Boots effect

DmytroTkachukPingleStudio:
## Motivation

Added Magic Boots effect  with scene for testing this effect.

## Summary of changes

- added Magic Boots effect
- added new scene for testing this effect

## How has this been tested?

Effect can be tested via test scene using cheat panel

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


BertovDev (changes requested):
Items VFX looking good so far!! and only we are having some troubles with this PR. 

The current `MagicBootsVFX` prefab is not working. After entering in the vfx graph and recompiling it we make it appear but it does not look as expected.

If we are missing something please let me know.


https://github.com/lambdaclass/curse_of_mirra/assets/62818001/64627623-9e05-4f69-b093-89f063093003

DmytroTkachukPingleStudio:
I've added missing assets, so now the effect should look as expected.

BertovDev:
> I've added missing assets, so now the effect should look as expected.

Nice! thank you.

BertovDev (approved):

PR #1685, opened 2024-04-17
Show items vfx usage for all players.

BertovDev:
## Motivation

Show Items to all players.

## Summary of changes

- Handle items for all players
- Handle items when Player is invisible. (We should thinks this better in a later iteration)

I would like to Iterate later on this in the following topics.

- Improve efficiency of the code. Maybe cache components and unnecessary calls.
- Returning in the correct time when the VFX is paused by a invisibility. 
- Maybe add a field to the proto message with an ID so we do not have to handle this in the client.

Ticket to this: https://github.com/lambdaclass/curse_of_mirra/issues/1686

## How has this been tested?

Pick items and use them.

## Test additions / changes

List tests added/updated.

## Checklist
  - [x] Tested in iOS.
  - [x] Tested in Android.


AngieDutra (approved):
Awesome. Tested in Android build

alfopisano (changes requested):
When testing on Unity desktop and iOS build I encountered some problems with the battle floor since it was not being displayed properly in any of them: 
img_7616_720

Also when testing on Unity desktop I could realize that some files are being unintentionally modified after starting a battle. To recreate this you need to start running the game, start a battle and then stopping the game remembering to save the changes afterwards. By doing so the following files are being modified: 
screenshot_2024-04-17_at_15 47 52_720_720

BertovDev:
> When testing on Unity desktop and iOS build I encountered some problems with the battle floor since it was not being displayed properly in any of them: img_7616_720
> 
> Also when testing on Unity desktop I could realize that some files are being unintentionally modified after starting a battle. To recreate this you need to start running the game, start a battle and then stopping the game remembering to save the changes afterwards. By doing so the following files are being modified: screenshot_2024-04-17_at_15 47 52_720_720

The white floor seems to be a problem of the build. Also no changes of what you commented had been added in this PR so it seems to be a "problem" already in main.

alfopisano (approved):
Nice!! Taking into account the problem mentioned is not from this PR I think it can be merged

Issue #1682, opened 2024-04-16
[FEATURE] Quests screen implementation


saugon:
## 🚀 Feature
This issue focuses on implementing the Quests screen in Unity utilizing the provided design. The goal is to incorporate the designed layout and functionalities:

- Tab system for daily and weekly quests.
- Current daily and weekly quests list.
- Quest card:
  - Quest description and reward.
  - Quest status.
  - Quest progression.
  - Action: claim.
- Rerolling feature:
  - When the player taps on reroll, the quests that can be rerolled will be highlighted, allowing the player to tap one of them to proceed with the quest roll. Once performed, the screen should return to its normal state.

BONUS:
- Add Quests link in the main screen:
  - Change the button text, now it says "Daily Quest", it should say "Quests"
  - Remove the "soon" badge
  - Add the hexagon above the button, with the corresponding number indicating the quantity of quests available to claim

jazimp:
Assets: https://drive.google.com/drive/folders/1ZpIXTO4d0sZbCJ8m62tDgiWJEX3Jlo8w?usp=drive_link
Buttons: https://drive.google.com/drive/folders/1h93h0kjVentkQGeoLqCXmc7HB0Y7pBGr?usp=drive_link
Background: https://drive.google.com/drive/folders/1Xw1OrKr1cw9cX-T7HgFbgeXygWx8hbPM?usp=drive_link
Hexagons: https://drive.google.com/drive/folders/1PpQOQ_HblQI8gKZnhLf8usDn_wvV0dir?usp=drive_link
Prototype: https://www.figma.com/file/qMZIYXmOCw8JXVEZ4m6nWu/Untitled?type=design&node-id=0%3A1&mode=design&t=TgMcJz5nkgEsjN7z-1

PR #1701, opened 2024-04-19
[GH-1682] Quests mockup

alfopisano:
Issue #1682 

## Motivation
This PR is the first part for the Quests screen. Here all components were created but none of them has functionality. That would be added on further iterations. 

## Summary of changes
Created the new Quests scene and its components. 

## How has this been tested?
Just go to the Quests section and take a look at it. 

The Quests screen: 
Screenshot 2024-04-24 at 16 37 43

The design: 
Screenshot 2024-04-19 at 12 48 47

-> Although the "REROLL" button looks different, checked with Jaz and it is the secondary button (the one I used) 

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


mgrunwaldt:
@alfopisano it could be nice if for PRs that involve UI you include a screenshot of what you made (this PR has this) and under it a screenshot of the original design, so it is easy to compare, plus the link to the original design

BertovDev (commented):
This is looking really nice and does this also includes the responsive design ? Because in the moment I change to others screen it looks broken.

alfopisano:
> This is looking really nice and does this also includes the responsive design ? Because in the moment I change to others screen it looks broken.

What do you mean with others screen?

alfopisano:
After talking to Gonza I will be making some changes to polish it ✨

AngieDutra (changes requested):
I love the slider idea in this implementation!
Implementations could be improved and would solve a lot of issues that show in the asstes and fonts. Also the assets config isn'r right. Take into account:
- Small assets are imported in 512 max size
- Button images are not set as sliced but as simple
- Avoid scaling images in only one value (x,y,z). If scaling changes should only be because a native size was applied and the scale should change in all axis the same amount to preserv resolution
- Ideally, images with point corners should be set to avoid braking on with or heigh change. Set those correctly or use the simple hack of setting to nativa size and rescaling.
- Use the copying reference hack I showed you to make sure the implementation is accurate to the reference
- Careful with text implementation. Check the extra settings margins are not modifies by accident
- There are no .5 values in font size. Try to set round values in the font sizes inputs

AngieDutra (changes requested):
Screenshot 2024-04-24 at 17 22 31

Talked with the design team and decided the following:
Quest should actually be quests
The soon label should be re added there until the hole section is done and available for users
Therefore, the link in this button to the quest section should be deleted also. To avoid having some "soon" labeled buttons that allow the user to enter a new scene and some that don't

AngieDutra (approved):
LGTM! Awesome PR

jazimp:
Good job! I'm leaving this ticket with adjustments to add when implementing functionality https://github.com/lambdaclass/curse_of_mirra/issues/1727

BertovDev (approved):

PR #1707, opened 2024-04-22
Setup sfx feedbacks

salvadorstrambini:
Checking Character Skills Sound effects
Proving new music

Explain why not how.

Adjusted the level of all characters skills SFX

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


BertovDev (changes requested):
Nicee great advance and be careful you pushed some unwanted changes.

- The TextMesh Pro folder should be in the ThridAssets folder in order to no be pushed to the branch.
-  The changes in the Resources/Fonts should not be pushed.
- The changes in  VFZ-Tools/.. also should not be pushed.

Remember to check what you push before doing it.

salvadorstrambini:
I've changed the lobby music in this same PR. Also added two sfx for H4ck skills

salvadorstrambini:
> Uma and valtimers SFXs are not set. Also I believe they need more volume. Wdyt?

You are right, i ve already fix it.

AngieDutra (changes requested):
I think you are missing the following metas to push:
Screenshot 2024-05-02 at 15 15 22

The overall volume of the sfx should be louder. To check this I has to mute the background music an set my phone to max volume. Ideally at max volume both music and sfx should be at the same volume. The the user can alter the level of each track in the settings.
Did an Android build and all 4 characters' sounds worked as expected. IMO the only thing left to do is adjust the initial volume values.
BTW my personal favorite is Valtimer's sfx, they totally portrait his power, great job!

AngieDutra:
In addition. Music and sfx aren't working in my Unity Desktop. Do they work for you guys? @salvadorstrambini @BertovDev

salvadorstrambini:
> In addition. Music and sfx aren't working in my Unity Desktop. Do they work for you guys? @salvadorstrambini @BertovDev

It works for me, but you need to open the settings for the sound to start.

alfopisano (commented):
Overall I think it works alright but I agree with the other comments, the volume of the skills (specially the ones from Muflus) should be louder. What do you think?

AngieDutra:
Agree, Muflu's are particularly low

BertovDev (changes requested):
Uma and H4ck basic seems to high in volumen. 

Also I believe we should reduce the 3DSound max distance.

AngieDutra:
@salvadorstrambini is adding new sounds to some skills

AngieDutra:
I noticed hack's basic improved, but I still can't hear muflus's skills as clear as the rest @salvadorstrambini

alfopisano (changes requested):
As discussed, I think Hack's skill's sfx are fine but we need to adjust the other ones (specially Muflus's ones) because hard to hear on top of the music.

tkz00 (changes requested):
The PR description should be updated to better explain to the reviewers what they need to pay attention to.

I played around a little in the branch and I found some issues:
- H4ck’s ultimate's sound finishes after the projectiles stop hitting the ground
- Muflus ultimate should have a wind up sound, for when he is “charging” the sound
- “Daniel Service” instead of “Denial of Service” in `client/Assets/Audio/H4ck/Daniel Service UP Leveled.wav`
- I think the audio files should have some kind of naming convention, it doesn’t need to be super strict
    - Why do some files have “4 6” in their names?
    - “Collect item 2” vs “Collect item 4”??

salvadorstrambini:
> * H4ck’s ultimate's sound finishes after the projectiles stop hitting the ground
> 
> * Muflus ultimate should have a wind up sound, for when he is “charging” the sound
> 
> * “Daniel Service” instead of “Denial of Service” in `client/Assets/Audio/H4ck/Daniel Service UP Leveled.wav`
> 
> * I think the audio files should have some kind of naming convention, it doesn’t need to be super strict
>   
>   * Why do some files have “4 6” in their names?
>   * “Collect item 2” vs “Collect item 4”??

4 6 is because of the date, for me to separate them from the previous files.

tkz00 (changes requested):
1. The Muflus Leap wind up sound is too quiet.

2. It looks like "Collect Item 2" & "Collect Item 4" sound files are included in the PR.
Are these currently being used in the game? If not, it might be best to remove them to keep the project clean.

3. #### File naming
- The naming convention for sound files could be improved for better readability.
Can we switch "UP/DOWN" to "start/stop" to match the SkillInfo fields?
- For instance, renaming H4ck's ultimate sound files to "H4ck_Denial_Start" and "H4ck_Denial_Stop" would be more descriptive.

4.  I noticed changes to `client/Assets/Prefabs/Camera/MainScreenCameraUI.prefab` Did you intentionally make any modifications there? When I checked out the file from the main branch, the PR seemed to function properly. Perhaps I missed something? If it was an misplaced changed then it should be reverted.

5. There are changes to both `CharacterBase.cs` and `Skill.cs`, the one on line 155 of `Skill.cs` can be reverted, and the original code can also be used for line 172 (`Sound3DManager sound3DManager = GetComponentInChildren<Sound3DManager>();`) with this change the modification made to `CharacterBase.cs` becomes useless, so it also should be fixed.

salvadorstrambini:
> 1. The Muflus Leap wind up sound is too quiet.
> 
>     2. It looks like "Collect Item 2" & "Collect Item 4" sound files are included in the PR.
>        Are these currently being used in the game? If not, it might be best to remove them to keep the project clean.
> 
>     3. #### File naming
> 
> 
>     * The naming convention for sound files could be improved for better readability.
>       Can we switch "UP/DOWN" to "start/stop" to match the SkillInfo fields?
> 
>     * For instance, renaming H4ck's ultimate sound files to "H4ck_Denial_Start" and "H4ck_Denial_Stop" would be more descriptive.
> 
> 
>     4. I noticed changes to `client/Assets/Prefabs/Camera/MainScreenCameraUI.prefab` Did you intentionally make any modifications there? When I checked out the file from the main branch, the PR seemed to function properly. Perhaps I missed something? If it was an misplaced changed then it should be reverted.
> 
>     5. There are changes to both `CharacterBase.cs` and `Skill.cs`, the one on line 155 of `Skill.cs` can be reverted, and the original code can also be used for line 172 (`Sound3DManager sound3DManager = GetComponentInChildren<Sound3DManager>();`) with this change the modification made to `CharacterBase.cs` becomes useless, so it also should be fixed.

1. The Muflus Up is too quiet in propouse because it needs to be that way
2. Collect Item is being used right now, i will check out
3. Yes, make sense 
4. I don t know about that
5. I don t know, we can see that in the office

PR #1711, opened 2024-04-23
Restore exit button

AngieDutra:
## Motivation

When the camera toggles was added the exit button got pushed out of the container

## Summary of changes

This PR adds that until a refactor on how we handle responsive lists is added. The complexity here is we have multiple horizontal and vertical layout groups that difficult the implementation of the content size component

When the team has a spare time we should look into this but it works for now

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.

### DIFF
--- a/client/Assets/Prefabs/Camera/CameraUI.prefab
+++ b/client/Assets/Prefabs/Camera/CameraUI.prefab
@@ -29139,6 +29139,26 @@ PrefabInstance:
       propertyPath: m_AnchoredPosition.x
       value: 0
       objectReference: {fileID: 0}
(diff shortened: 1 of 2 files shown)


BertovDev (commented):
Why do you have changes in CameraUI prefab? it is not only necessary to change the BattleSettingsPrefab?

AngieDutra:
I created the [issue](https://github.com/lambdaclass/curse_of_mirra/issues/1713) also

BertovDev:
I checked out the changes and your solution is still working.

AngieDutra:
I did the same but they didn't in my case 🤔

PR #1717, opened 2024-04-24
General performance improvement

### LINKED ISSUE #470: Unity performance
The main goal is to improve the game performance. To do that, we first need to collect some metrics and once we know where the pain points are, we can plan how to address the improvements.

#### We need to measure the impact of:
- Scripts
- Assets size/compression.
- Models
- Visual effects
- Shaders

#### Output
- The results need to include a maximum boundary for the weight of 3D assets, measured in poly count. Context: 3d modelers need to know the maximum poly count they are allowed to use for each character. Therefore, we need to estimate the maximum poly count we can handle and then estimate how much we will use for the complete map and for each of the 20 characters in the match.

saugon:
Closes #470 

## Motivation
The game isn't running well enough on low-end devices.

## Summary of changes
Core:
- Replaced Linq functions on frequently-updated code
- Refactored code to improve memory management
  - Battle.cs
  - CharacterFeedbacks.cs
  - EventsBuffer.cs
- Changed map realtime light to baked
- Tweaked materials (changed shaders to use "Simple lit" insted of Lit)

Aditional:
- Fixed character damage feedback

## How has this been tested?

The game should run faster than main branch without having any visual changes.

## Test additions / changes

List tests added/updated.

## Checklist
  - [x] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/EventsBuffer.cs
```diff
@@ -93,24 +104,21 @@ public bool playerIsMoving(ulong playerId, long pastTime)
...
+                return true;
+            }
+            serverPlayerUpdate = new Entity(followingEventToRender.Players[playerId]);
+            if (serverPlayerUpdate.IsMoving)
+            {
+                return true;
+            }
         }
 
         return count >= 1;
```
BertovDev, line 124: `count` is not being use anymore, this could be `return false` instead.

BertovDev (approved):
Awesome, some things I will do in another PR to not stop this to being merge.

- Make shadow de-active on death
- Make shadow dissapear on invisibility 
- Maybe change shadow material to `SimpleLit` ?
@saugon

client/Assets/Scripts/EventsBuffer.cs
saugon: Fixed shadow handeling!

> Maybe change shadow material to SimpleLit ?

This would be correct if it were a Lit material, but in this case, it's Unlit.

BertovDev (approved):

AngieDutra (approved):
Awesome!
Tested in android build:
- FPS: 110 - 119
- PING: 35

To iterate:
- The shadows imo look a bit off world. We could improve how they look later on
- The death splash lights, specially with h4ck, looks a bit dim but look better than a few commits earlier

Overall great PR ✨

PR #1765, opened 2024-05-15
Fix Dashes animation getting stuck.

BertovDev:
## Motivation
Dash animations were getting stuck sometimes. 
This was happening due to the duration of the animation being longer on the client than on the backend. We need to fix this. 
Another problem we are having is that we are handling two different skill animations that are very similar in two different ways. This are the dash and ulti animations from Muflus and Uma. We need to make the dashes animation use the same method as the ulti. By doing this we will not only follow a same criteria but also fix the problem mentioned above 🙌

## Summary of changes
- Modified Muflus and Uma dashes scriptable objects to include the three animations than compone the skill into the AnimationsList it has (this is what is being done on the ulti). 
- Added a speed multiplier to accelerate the animation to the desired speed. 

## How has this been tested?
Test Uma and Muflus dashes. To force the animation to get stuck you can change in the backend the duration of the dash for the characters. When making it smaller the animation would fail multiple times. This should not happen on this branch. 

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


tkz00 (changes requested):
https://github.com/lambdaclass/champions_of_mirra/assets/67983386/17e456d9-f7c3-4af8-9c28-322d9c186d82

I had the animation getting stuck problem playing with Uma. To test better I set the cooldown of the skill to 1000ms in the config, don't know if that could've affected the outcome, I think not since the cooldown is longer than the duration of the skill itself (I don't even know if it being shorter would break something). But maybe the problem is with other states conflicting in the animator, such as the basic attack skill parameters maybe messing it up.

tkz00 (changes requested):
I still have the issue 😢 


https://github.com/lambdaclass/champions_of_mirra/assets/67983386/1ec41eeb-e365-45da-99d7-7a457a154a26


https://github.com/lambdaclass/champions_of_mirra/assets/67983386/7ca8beb0-03d9-41d6-ab0c-ccba712642dd

tkz00 (approved):
Works great! Tested on my android device!

AngieDutra (changes requested):
Correct me if I'm wrong, but I think we are not using the backend execution duration value from the config to affect the duration of the animation. I believe this should be the case and the sum of each animation clip that is in the dash, regardless of how many they are, should represent the total of the backend value.

AngieDutra (changes requested):
Now I have new issues with the dashes (Muflus's)
With the dash as it is and the main config I guess we need to adjust the percentage for each animation since the roll happens quicker than the roll in and the roll out.

The following vid shows what happens when I speed up the execution speed to 30


https://github.com/lambdaclass/champions_of_mirra/assets/82987608/40b141c1-993b-46f3-a33f-7d2290e3af06

AngieDutra:
Here is a vid of this PR using the main config:
https://github.com/lambdaclass/champions_of_mirra/assets/82987608/3865d4f7-f6d1-4768-9d8f-05e76dc71ea0

alfopisano:
I agree with the percentages adjustment, but the Issue with a speed of 30 could be an actual problem when on prod?

AngieDutra:
It is but I though it was part of the scope for this PR. If not you can totally dismiss the last part

AngieDutra (changes requested):
Muflus is working awesome! I just got a bug with uma:
main:

https://github.com/lambdaclass/champions_of_mirra/assets/82987608/8bd8c95e-c19e-446e-8cc1-59aa49b763d6


This branch:


https://github.com/lambdaclass/champions_of_mirra/assets/82987608/a746443e-6b79-42a6-b130-b1f3fa18ccd7

alfopisano:
Tried to recreate the Issue but I couldn't, this was recorded on an iOS build: 
https://github.com/lambdaclass/champions_of_mirra/assets/116388676/17fed160-994c-42e0-b039-52e1dce2801d
Did you do anything special to have this problem?

AngieDutra (commented):
Just tested along side @alfopisano this issue. We had the following results:
in main branch:
- Sometimes the movement froze but not the animation
- Sometimes the animation froze but not the movement
in this PR:
- Sometimes the movement freezes but not the animation
- The animation freezing issue is not happening here
I have tested in desktop and Android build. Maybe it could be best to iterate since the main issues are fixed and take care of this in a separate PR. I'll ask the team and come back with an update on this. I see no other issues

AngieDutra (approved):
The team decided to iterate
Tested on desktop and Android build
Awesome work here

tkz00 (approved):
Looks great! I tested it on my android device!

PR #1770, opened 2024-05-15
Fix unwanted file changes

BertovDev:
## Motivation

- Text mesh pro problems.

TextMeshPro dynamic font assets have a very annoying habit of saving their dynamically generated binary data in the same text file as their configuration data. This causes massive headaches for version control.

- Danger zone problem

Renderer.sharedMaterial returns the actual instance(s) of the Material at renderer.sharedMaterials[0] that is on the object.
This can be modified, and it will affect every use of this Material in the Scene.

renderer.material, on the other hand, creates a copy of renderer.sharedMaterials[0]

So with this we are not overriding the material saved in the scene.

## Summary of changes

- Prevent dynamics fonts changes to be saved
- Use material instead of shaderMaterial to do changes.
- Set rotation after instantiating the model Denial_AOE.
- Regenerate each font again as Static.

## How has this been tested?

Normal play the game, no changes of fonts and the danger zone should appear in the source control.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


alfopisano (changes requested):
Restores any changes from the terminal. Played on unity desktop, after the battle ended I stopped the game and saved any possible changes. When heading back to the terminal this files were modified: 
Screenshot 2024-05-15 at 16 27 09

alfopisano (changes requested):
Tested as described previously and encountered the following problems: 
- This file was modified:
Screenshot 2024-05-15 at 18 23 38

- I asume it has to do with the change above but some font were looking weird:
Screenshot 2024-05-15 at 18 22 00
Screenshot 2024-05-15 at 18 21 34

BertovDev:
> This file was modified:

Did you restart unity?

alfopisano:
Tried restarting unity and the problem persisted.

AngieDutra (changes requested):
In addition, I'm having the same issue. Pink means missing material, maybe you forgot to save something in the fix process?
Screenshot 2024-05-20 at 11 48 44

AngieDutra:
Or maybe you need to replace the fonts one by one through the game?

tkz00:
This PR is trying to solve two different issues, the fonts one is being handled here: [Fix dynamic font assets changing version control](https://github.com/lambdaclass/champions_of_mirra/pull/1814), the materials/textures one will be investigated and tackled later on.

PR #1793, opened 2024-05-23
[GH-1792] Add Kenzu Character

tkz00:
## Motivation

The objective for this PR is for the VFX's team to be able to reference Kenzu's model and animations to start working on it's effects.

## Summary of changes

- Kenzu's model, textures and animations have been imported into the project in their respective folders.
- An animator controller for Kenzu has been added to the project, with a state for every respective animation, the necessary transitions and parameters.
- A Kenzu character prefab was added with it's model and animator controller.
- A new scene `VFX_Kenzu_test_scene` was added to test all the animations.

To check out the animations you have to enter the previously referenced scene on play mode, in it's hierarchy select the animator inside the `Kenzu` model in `VFX_Kenzu_test_scene/Kenzu/CharacterModelAnchor/`, with the animator screen open you can tweak the parameters to trigger each animation. Keep in mind there are duplicated states for the basic attack, block and pounce animations, since our backend and client don't support these "combo" styled skills we added a way to try these animations, for this use the `Basic_Test` and `Blocking_Test` parameters, while the `Walking` and `Skill3` work for both the test and in game functionality.

For the basic attacks the idea is to have a 3 part combo where the attacks are sequenced, while it can be exited after the first or the second attack without performing the remaining part of the combo, you can't execute the second or third attack without performing the previous ones first. Toggling `Basic_Test` will take you through the full 3 part combo.

The blocking/pounce is slightly more complicated, there is an initial start block animation performed when turning on `Blocking_Test`, where the character raises it's weapon, after it's finished it keeps the weapon up and enters and enters a blocking in place or a block while walking animation depending on the `Walking` parameter. Turning `Blocking_Test` off represents another character hitting Kenzu, which when struck during the blocking stance performs a pounce on the attacking enemy.


tkz00:
Since we aren't currently going forward with Kenzu's implementation this PR will be closed, the branch will still be active to continue with the task in the future.

Issue #1617, opened 2024-03-21
[FEATURE] Displaying connectivity Icon for connection issues 


saugon:
## 🚀 Feature
In this task, we aim to show a connectivity icon on the screen to alert players when they encounter connectivity issues during gameplay. It involves researching the criteria for determining when a connection is unstable and identifying optimal moments to trigger the icon.

We can use this [icon](https://github.com/lambdaclass/curse_of_mirra/assets/6811225/81d5eb75-367f-4934-9b50-c9fc06e96bdf) in the meantime. Related to #1614

jazimp:
https://drive.google.com/drive/folders/1JvzwC4_Blk11WDn96eKysPKZkIBBY9l_?usp=drive_link reference, asset and animation reference

PR #1798, opened 2024-05-29
[GH-1617] connectivity issues feedback

Nico-Sanchez:
> [!WARNING]  
> This has to be merged with https://github.com/lambdaclass/mirra_backend/pull/619

Works towards solving #1617 

## Motivation

When a user has connectivity issues, feedback communicating this should be returned to the user.

## Summary of changes

- An algorithm measuring jitter (high variations of ping between messages) and the time elapsed since the last update were implemented.
- When any of these two cases are met, a bad connection icon will be display on the center of the screen.
- For the icon to disappear the connection must be stabilized.
- If the client receives no messages for 10 seconds, then it will return to the main screen with a disconnection message popup.

All the parameters used in the algorithms to detect the stability can be set in the backend file `client_config.json`.

### To test the complete lack of messages

This won't work if you are testing on the Unity editor since to disconnect the computer to the internet or kill the backend running locally you have to loose focus of the application, and by the way Unity handles this the icon will appear, even if no connection issues arise, tainting the results of the test.

- Enter a match and in the middle of it disconnect your device from the backend. You should see the icon popup after 3 seconds without internet, then after 10 seconds the game should take you to the main screen with a popup explaining what happened.

### To test jittering

- You'll have to try this al least once with a good connection to see that the icon doesn't appear.
- After that try to play a match with a bad connection, see that when the match itself feels laggy the icon is displayed, and when the connection is stable again, and you can play correctly, the icon disappears. It's best to pay attention to the bots in the game to see if your connection is good, since your own character has client prediction so it might move correctly even with a bad connection.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


client/Assets/Scripts/ConnectionHealthAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        if(timestampDifferences.Count >= GameServerConnectionManager.Instance.timestampDifferenceSamplesToCheckWarning)
+        {
+            if(timestampDifferences.Max() - timestampDifferences.Take(GameServerConnectionManager.Instance.timestampDifferenceSamplesToCheckWarning).Average() > GameServerConnectionManager.Instance.showWarningThreshold)
+            {
+                unstableConnection = true;
+            }
+            else if(timestampDifferences.Max() - timestampDifferences.Average() < GameServerConnectionManager.Instance.stopWarningThreshold)
+            {
+                unstableConnection = false;
```
manucamejo, line 64: You are calculating the same things several times: `timestampDifferences.Max()`
I also think that we should separate the algorithm to activate and deactivate the icon because they are things that we should check with different time intervals.

manucamejo (changes requested):

client/Assets/Scripts/ConnectionHealthAnalyzer.cs
```diff
@@ -0,0 +1,81 @@
...
+        long timestampDifference = newTimestamp - lastUpdateTimestamp;
+        lastUpdateTimestamp = newTimestamp;
+
+        if(timestampDifferences.Count > GameServerConnectionManager.Instance.timestampDifferencesSamplesMaxLength)
+        {
+            timestampDifferences.RemoveAt(0);
+        }
+        timestampDifferences.Add(timestampDifference);
+
+        GameServerConnectionManager.Instance.currentPing = (uint)timestampDifferences.Last();
```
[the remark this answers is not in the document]
Nico-Sanchez: We don't really need to do that. It was for testing purposes. See that it's changing the currentPing and it shouldn't.
I deleted it in [Restore ping shown in UI](https://github.com/lambdaclass/champions_of_mirra/pull/1798/commits/be7ab2e0eb1268cb3d89e8aebe584f72aec3dcea)
Seems like it got pushed again in future commits. Will delete it again.

client/Assets/Scripts/ConnectionHealthAnalyzer.cs
Nico-Sanchez: Done the first one.
About the 2nd one, we can totally do that. However, I'd like to know the motivation behind that change since its working fine now.
I believe you mentioned something about performance in our huddle. Do we have any tool to measure the algorithm performance?

tkz00: We discussed this in person with Nico and Manu, different methods for detecting spikes in latency vs checking if the connection is stable have been implemented to encapsulate behavior that previously was inside OnGameEventTimestampChanged, but it was agreed that a reduced frequency to check if the connection is stable, and by extension, the disappearance of the icon, would be left for a future iteration of the feature.

alfopisano (approved):
Nice!! Tested as described on an iOS build and it worked as expected 🙌

alfopisano (dismissed):
Tested again on iOS build and everything worked fine!!

client/Assets/Scripts/ConnectionHealthAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        GameServerConnectionManager.OnGameEventTimestampChanged += OnGameEventTimestampChanged;
+        GameServerConnectionManager.OnMatchFinished += OnMatchFinished;
+    }
+
+    void Update()
+    {
+        if(!matchFinished && lastUpdateTimestamp > 0)
+        {
+            long msSinceLastUpdate = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds() - lastUpdateTimestamp;
```
manucamejo, line 24: You should move this variable to the top of the file, we are creating a new variable in each update.

client/Assets/Scripts/ConnectionHealthAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+        if(timestampDifferences.Count < GameServerConnectionManager.Instance.timestampDifferenceSamplesToCheckWarning)
+        {
+            return;
+        }
+
+        if(SpikeDetected(timestampDifferences))
+        {
+            unstableConnection = true;
+        }
+        else if(IsConnectionStable(timestampDifferences))
```
manucamejo, line 64: This functionality should run every second instead of on each update. But don't change it now. Can you create an issue for that?

manucamejo (changes requested):

manucamejo (approved):

alfopisano (approved):

PR #1814, opened 2024-06-04
Fix dynamic font assets changing version control

tkz00:
## Motivation

A quality of life improvement to avoid TMP's dynamic font assets to change and be registered by version control on every new character appereance.

## Summary of changes

Added a static script that runs every time the Unity editor tries to save serialized assets or scenes, then after verifying it's a font assets clears all the font asset data, thus removing the changes to the asset file.

## How has this been tested?

Try to add new UI texts in the game with the dynamic fonts of the project (the ones commited in this PR), using weird or unused characters, then save the scene, play the game through that scene just to be sure, and after exiting play mode, check if git has registered changes on the font's asset files.

References:
https://github.com/STARasGAMES/tmpro-dynamic-data-cleaner/tree/main
https://forum.unity.com/threads/tmpro-dynamic-font-asset-constantly-changes-in-source-control.1227831
https://docs.unity3d.com/ScriptReference/AssetModificationProcessor.html

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


alfopisano (commented):
Nice!! It was really annoying to have those changes anytime you want to commit something 🙌. 
I just have a question, the problem with the file `client/Assets/VFX-Tools/PIngle_VFX/DamageZone/M_SandFlow.mat` will be addressed on another PR or should also be fixed here? Because that file is also being modified everytime you play.

tkz00:
> Nice!! It was really annoying to have those changes anytime you want to commit something 🙌. I just have a question, the problem with the file `client/Assets/VFX-Tools/PIngle_VFX/DamageZone/M_SandFlow.mat` will be addressed on another PR or should also be fixed here? Because that file is also being modified everytime you play.

There will be another different PR for that, even though the issue is the same, assets getting changes registered by version control, the cause for the fonts and the materials is different.

alfopisano (approved):
Nice!!

ncontinanza (approved):
Nice job!

PR #1816, opened 2024-06-04
Fix spectator mode

tkz00:
## Motivation

Sometimes after being killed the spectator just stood still instead of following the killer player, or a random player in case that you where killed by the zone.

## Summary of changes

Removed lines 111 to 115 of `KillFeedManager.cs` which assigned id `9999`, the id we use for the zone, as the killer id when the player's health reached 0 or less health with no kill event for the player received by the client, the logic of this differed to the one we have for normal kills from an opponent, which were only registered via kill events. This discrepancy caused the spectating behavior of being killed by the zone (which was to follow no player and stay still, for some reason) to be used sometimes even when killed by another player.

Now the spectating logic of being killed by the zone is to follow a random player, and it is never used when killed by an opponent.

Also the spectating overlay has been changed, the SPECTATE button removed, the CONTINUE button changed to QUIT, and the red overlay has been made a little more transparent. In addition to this, some typos and function names that didn't follow the naming conventions have been fixed.

## How has this been tested?

- When killed by the zone you should follow the player with the ID of the first player that comes from the backend.
- When killed by an opponent your camera should follow his character from now on.

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.

### DIFF
--- a/client/Assets/Scenes/Battle.unity
+++ b/client/Assets/Scenes/Battle.unity
@@ -982,16 +982,6 @@ PrefabInstance:
       propertyPath: ButtonReleased.m_PersistentCalls.m_Calls.Array.data[0].m_Arguments.m_ObjectArgumentAssemblyTypeName
       value: UnityEngine.Object, UnityEngine
       objectReference: {fileID: 0}
-    - target: {fileID: 5513555720094977976, guid: 9edcf7e494cf3e640b18cb7495fb179a,
-        type: 3}
-      propertyPath: m_Color.a
-      value: 0.3529412
-      objectReference: {fileID: 0}
-    - target: {fileID: 5624234333185236875, guid: 9edcf7e494cf3e640b18cb7495fb179a,
-        type: 3}
-      propertyPath: m_text
-      value: Quit
-      objectReference: {fileID: 0}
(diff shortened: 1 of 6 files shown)


client/Assets/Scenes/Battle.unity
alfopisano, line 1: Why were this changes made on the scene and not on the prefab?

client/Assets/Textures/Uma.meta
alfopisano, line 1: Why has this file been changed?

tkz00: The asset slipped through, I need to pay more attention, already removed.

client/Assets/Scenes/Battle.unity
tkz00: Right! I'm not used to the way the project is structured, but I have to adapt to follow the project's conventions. The fix is already made, the changes have been applied only in `DeathSplash.prefab`.

alfopisano: Great!! 🙌

client/Assets/Textures/Uma.meta
alfopisano: Oh it happens to us all 😄

alfopisano (approved):
Nice!! Tested on my iPhone and it worked great!!

ncontinanza (approved):
LGTM!

PR #1817, opened 2024-06-05
[VFX] Reworked Damage zone effect

DmytroTkachukPingleStudio:
## Motivation

Reworked Damage zone effect. Now it has additional conus-like shape to show exect effect area and additional shadow particle to highlight main sand effect.

## Summary of changes

Added new Renderers to SafeZone.cs to control _AlphaMultiplier and _Progress parameters. Also added new objects to existing effect.

## How has this been tested?

New effect can be tested via test scene VFX_DamageZone_test_Scene.unity using DamageZoneManager.cs ore via regular gameplay

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


AngieDutra (changes requested):
Awesome! I have a few comments:
Screenshot 2024-06-14 at 18 51 31
Screenshot 2024-06-14 at 18 51 44
Screenshot 2024-06-14 at 18 51 55
- The zone is starting before it should
- When the character or the boxes are in the zone they get lost. I don't know if the zone was intended to be this solid but as a player i believe it would be better if it wasn't as much
- There is no transition for when the zone "cuts" through a collider. This kind of cut in the zone takes me out of the illusion of a sand storm

DmytroTkachukPingleStudio:
Added optimized Sand Storm (ground sand) effect.

DmytroTkachukPingleStudio:
Added more optimizations on the shader and disabled the shadow.

PR #1821, opened 2024-06-05
Refactor pool's and add buff singularity feedback for Valtimer

### LINKED ISSUE #1820: [FEATURE] Visual feedback for Valtimer's basic colliding his ultimate
## 🚀 Feature

When Valtimer's basic attack collides with his ultimate skill, the latter's pull force grows and so does it's duration. This is already handled in the backend ([[gh-460] Add effects to entities on projectile collision](https://github.com/lambdaclass/mirra_backend/pull/463)), but the respective visual feedback is still not implemented on the client.

To do this, the backend is now sending the necessary information for the client to make this feedback, when appropriate, the VFX for the ultimate skill should have some of it's properties modified.

agustinesco:
Closes #1820 

> [!WARNING]  
> Works with backend branch [Add pool effects to messages.proto](https://github.com/lambdaclass/mirra_backend/pull/685)

## Motivation

Valtimer's ultimate, SIngularity, get's buffed when hit by it's basic attack, currently, there isn't visual feedback for this on the client.

## Summary of changes

- I've refactored how pool skills are handled on the client, previously they were just VFX's tied to the ability that spawned when cast, based on player actions. This has been changed to pools spawning based on the pool entities that come from the backend, very similar to how projectiles are handled.
- In addition to this, now pool states that come from the backend can have `effects`, the same as players. Only for the effect `buff_singularity`, a color an size change has been added to Valtimer's ultimate, in the backend this effect is applied when Valtimer's basic hit's it's ultimate.

With the refactor made it should be easy to add new feedback for any pool skill having a new effect.

DISCLAIMER: Visual effect parameters are hardcoded in the client, they aren't taken from the backend, for Valtimer's `buff_singularity` the only value relevant for this is the duration increase of the pool, currently is 1s in the backend, but if it's changed, that won't be automatically reflected in the client. For this to change some new values should be sent by the backend.

## How has this been tested?

In a match with Valtimer as your character, hit your ultimate with your basic ability, a momentary change of color and diameter should happen, also the duration of the effect should be extended by 1 second.

Video for reference:

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/9208d3bf-41f0-42ba-80e0-2975f09fc7ff

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


alfopisano (dismissed):
Nice!! Tested on my iOS device and the ulti + basic worked great!!

alfopisano (dismissed):
Nice!! Tested on my iOS device and worked really nice

client/Assets/Scenes/Battle.unity
AngieDutra, line 1: Why did you change battle scene?

tkz00: You are right, but due to time constraints this issue will be ignored and fixed later, this is registered in: [[FIX] Move pools handler from Battle scene to BattleManager prefab](https://github.com/lambdaclass/champions_of_mirra/issues/1904)

tkz00: Since I had to make other changes this has been fixed and tha battle scece is no longer modified in this branch.

alfopisano (approved):
Great!! Tested on my iOS device and worked amazing

PR #1822, opened 2024-06-06
[VFX] Added Giant Fruit item effect

DmytroTkachukPingleStudio:
## Motivation

Added the Giant Fruit item effect along with its scriptable objects and test scene. The EffectCharacterMaterialController.cs script contains the necessary test code to test the effect in the test scene.

## Summary of changes

Added Giant Fruit item effect, scriptable objects with settings for the effect and test scene with cheat panel.

## How has this been tested?

Effect can be tested via cheat panel in Itens test scene

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


DmytroTkachukPingleStudio:
I just added a separate script for scale control, and now it works in all scenes. It has a default duration of 5 seconds (I'm not sure what the exact value should be), but it can be adjusted additionally during spawning or any other way if necessary.

PR #1844, opened 2024-06-13
Crate model update

### LINKED ISSUE #1843: [FEATURE] Update crate model
## 🚀 Feature
Update the existing chest model to improve its visual quality and add animations for hitting and opening the chest.

saugon:
Closes #1843 

## Motivation
We now have a 3D asset for the chest, which can replace the current placeholder. Additionally, we have a shake animation for when a chest is hit and an opening animation for when it is “destroyed.”

I’ve also refactored the implementation to improve its organization and clarity, since it was a bit messy. 

Observation: we should add a VFX and/or animation for when a chest disappears, but this is out of the scope of the current issue.

## Summary of changes
- Replaced the 3D model.
- Created the model material.
- Created an animation controller to use the new animations.
- Refactored all the scripts related to Crates login.

## How has this been tested?
Playing a match:
- Chests should have the new model.
- They should trigger a shake animation when they are hit
- They should trigger an open animation when they are destroyed.

## Reference
https://github.com/lambdaclass/champions_of_mirra/assets/6811225/55a80e9c-1750-447a-9900-30c0b0fbf4db

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


alfopisano (approved):
Nice!! Tested on iOS build and worked great

tkz00 (approved):
Great addition, I feel there are some things that still need polish like the crates animations concatenating but aside from that looks great.

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/bef6a9bd-cf9a-4dfc-b681-14fb47552089

See for example that if two of H4ck's projectiles hit at the same time the crate performs the hit animation twice, one after the other, or that it reproduces hit animations before playing the opening one.

PR #1847, opened 2024-06-14
Fix JWT issues when changing servers

AminArria:
## Motivation

Currently in a client if you get a JWT for one server and then change servers the client will try to re-use that same JWT. In the case of Brazil and Europe it would work cause currently they share gateway server, but if you go from localhost to Europe then you face problems. Essentially when you try to go across gateway servers and have a JWT in your client 

## Summary of changes

- Extend `SelectServerIP` to change the gateway server to use and the PlayerPrefs key where the JWT for that server is stored
  - Now `CUSTOM` expects 2 IPs, arena and server in a format `arena_ip gateway_ip` (notice the space, also gateway needs the http/https part)
- Extend `ServerUtils` to use the new information in `SelectServerIP`
- Small cleanup of unneeded `IP` prefabs

## How has this been tested?

To test this play across servers without clearing PlayerPrefs or app data

To test with Ngrok you will need to extend your configuration to be able to handle both tunnels. You can modify you ngrok config with `ngrok config edit` and putting
```yaml
tunnels:
  arena:
    proto: http
    addr: 4000
  gateway:
    proto: http
    addr: 4001
```

And then you can do `ngrok start --all`

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.

### DIFF
--- a/client/Assets/Prefabs/ServersOptions.prefab
+++ b/client/Assets/Prefabs/ServersOptions.prefab
@@ -34,7 +34,6 @@ RectTransform:
   m_ConstrainProportionsScale: 0
   m_Children:
   - {fileID: 818923841214956151}
-  - {fileID: 1099233914732415609}
   m_Father: {fileID: 4506445334052334876}
   m_RootOrder: 2
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
@@ -174,7 +173,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 61d7643ed2a634eec8f60a3be5d86efa, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
-  IP: {fileID: 569503291951164151}
+  IP: {fileID: 0}
   serverName: {fileID: 5690564450626468402}
 --- !u!1 &103087406830891457
 GameObject:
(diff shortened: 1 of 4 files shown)


tkz00 (changes requested):
https://github.com/lambdaclass/champions_of_mirra/assets/67983386/aa477ad4-7e63-4a94-a0fe-cbe4cd7f15b8

I can't go through the title screen when setting the custom server, but when using the Brazil one it works fine.

ncontinanza:
> muted.mp4 
> I can't go through the title screen when setting the custom server, but when using the Brazil one it works fine.

@AminArria please explain in the log that the input should be `ARENA_IP https://GATEWAY_IP`

ncontinanza (approved):

tkz00:
> muted.mp4
> 
> I can't go through the title screen when setting the custom server, but when using the Brazil one it works fine.

This was because `https://` was missing from the second URL

tkz00 (changes requested):
https://github.com/lambdaclass/champions_of_mirra/assets/67983386/c3982a96-6876-498f-a1e8-cbcd2e4e47b8

Screenshot 2024-06-14 at 5 47 27 PM


I can't go past the title screen now, sometimes I go through to the mainscreen but can't enter a match, and sometimes I can play matches, but generally it doesn't work.

tkz00 (approved):
Great, works good on my android device. For anyone reading this, I don't know if it has something to do with having multiple ngrok tunnels, or something related to my phone, but it takes me several minutes to login from the title screen, after that you can play normally, but just a disclaimer, since I believed this PR didn't work because of this.

PR #1856, opened 2024-06-18
Improved performance on damage feedback

### LINKED ISSUE #1855: [BUG] H4ck's ultimate crashes the game
## 🐛 Bug Report
The game freezes or crashes when executing H4ck’s ultimate skill on some low-end devices.

saugon:
> ℹ️  This PR was initially created to address issue #1855. However, due to multiple factors contributing to performance issues with H4ck’s ultimate skill, and the primary issue being related to the VFX, this PR has now been updated to focus on general performance improvements.

## Motivation
We detected some performance issues on some low-end devices when H4ck's ultimate was triggered. The main issue was an incorrect haptic triggering that was running multiple times. 
Additionally, I refactored the color overlay implementation. It’s now cleaner, more lightweight, and optimized, though the performance impact isn’t as significant as the haptic fix. 

## Summary of changes
- Removed unnecessary haptic feedback calls.
- Updated character shader to support damage overlay color.
- Refactored damage overlay implementation.

## How has this been tested?
- Tested with a low end device (which had significant FPS drops when the skill was used before the fix) 
- Playing with H4ck and using its ultimate skill on multiple targets.

The performance should not be affected by the skill.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


AngieDutra:
I've tested two Android devices with @tkz00 and they still crashed when using two hack ultis at the same time. I will keep on testing with an IOS build as well

saugon:
I’ll change the PR to draft. Tomorrow, I’ll try to use a phone that still has issues. If I can’t, we can still merge this PR because it also improves the performance in general.

AngieDutra:
Talked with @saugon how to review this best:
- Using hack's ulti he saw a drop of frames which in this PR happens no longer
- In addition he used a config to span the bots in the middle of the map
- Using ngrok fixed in this [PR](https://github.com/lambdaclass/champions_of_mirra/pull/1847) already in main you can play with that config and damage more players with this skill and further review this drop of frames improvement

AngieDutra (commented):
Tested on Android and IOS. I couldn't see a big difference on either device and I couldn't test on a lower end Android. With that being said I was able to play well in both devices. IOS was a bit worse but this happened in main too.

tkz00 (changes requested):
- The haptic feedback for damaging an enemy or crate is completely gone in my android phone (Samsung S10+)
- Also it didn't get any better in respect to performance, but I don't know if my problem is the one being targeted in this PR.


https://github.com/lambdaclass/champions_of_mirra/assets/67983386/17d83ea4-889c-46e8-a861-73c43599b38a

saugon:
> * The haptic feedback for damaging an enemy or crate is completely gone in my android phone (Samsung S10+)

Good catch, i will fix that, thanks!

>  * Also it didn't get any better in respect to performance, but I don't know if my problem is the one being targeted in this PR.

That specific issue is related to the VFX and being worked on in another other PR.

tkz00 (commented):
The code reorganization is a great improvement, much easier to follow. That said, I'm not sure I can see a performance difference between the current state of the game and this branch, based on the fps metric on top of the screen, both still have my screen flickering issue, but it has been explained that this PR doesn't aim at solving this issue. What do you think?

**This branch build:**

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/1596d1b0-d736-4008-8ff5-921b12dbb76a



**Current state on play store:**

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/7e176050-dd04-4cb6-8e9f-eae09fbd0030

saugon:
> The code reorganization is a great improvement, much easier to follow, also the logic to avoid running the haptic feedback multiple times works great. That said, I'm not sure I can see a performance difference between the current state of the game and this branch, based on the fps metric on top of the screen, both still have my screen flickering issue, but it has been explained that this PR doesn't aim at solving this issue. What do you think?

The difference could be seen in two ways:
- In the profiler: doing a deep profile or custom profile marks and compare the time spent in that function (unless you get a very small measurement like 0, which indicates the device can easily handle it).
- In low-end devices.

@AngieDutra has already measured the performance improvement (though she hasn’t commented here yet). If this is blocking you due to the devices you have access to, don’t worry about it and let’s continue with the review. This PR is blocking others that rely on changes made in the character prefabs.

> These are some comparisons to show how many times the haptic feedback has been executed in the same situation using prints.

@tkz00 Nice test! I already noticed this issue, but we have to wait until Manu’s update on player movement and actions is merged. Once that’s done, we can trigger the feedback only if there has been a change, instead of on every frame.

AngieDutra:
As @saugon said, I reviewed the performance improvements yesterday in a Redmi Android device. I'll review asap the last additions and use the profiler this time to check further

AngieDutra (approved):
Tested on desktop and Android build. It seems to be working as expected

tkz00 (approved):
Great! Looks good, works on my android device. The haptic feedback calls have drastically decreased, although I haven't been able to profile exactly how much of an impact that makes on the performance of the game, seeing a lower number of calls to the vibration is enough.

PR #1858, opened 2024-06-19
[VFX] H4ck Ultimate freeze fix

DmytroTkachukPingleStudio:
## Motivation

H4ck Ultimate compiles shaders during gameplay instead of the loading screen. I've added this VFX to the Prewarmer list, so now it will compile all necessary shaders under the loading screen.

## Summary of changes

Added H4ck Ultimate to Prewarmer list.

## How has this been tested?

Fix can be tested via regular gameplay

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


AngieDutra:
I tested with IOS device and after a while the image froze for me. However, my connection was okey and I could still use my movement input since I was able to move and the other player could see me do so

tkz00:
On my android device, I have the same problems as always, screen flickering when H4ck's ultimate appears and the application crashing when the match ends.

AngieDutra:
This is done in another [PR](https://github.com/lambdaclass/champions_of_mirra/pull/1878)

PR #1865, opened 2024-06-26
[PERF] Refactored character rotation

saugon:
## Motivation
This change addresses a tiny performance improvement by eliminating unnecessary physics calculations made by the CharacterOrientation3D component from TopDownEngine. I replaced this component and its dependencies with a simpler in-house solution, enhancing performance and reducing unused components.

### Performance impact
Not that much, only ~5% of the BehaviorUpdate (though it’s in the top 5 time-consuming updates).

## Summary of changes
- Removed TopDownController3D and CharacterOrientation3D components from character prefabs.
- Refactored rotation and aiming direction functions that were using CharacterOrientation3D component.

## How has this been tested?
Playing with all the characters and verifying their rotation while moving and executing skills.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


alfopisano (dismissed):
Played with all the characters on an iOS build and the rotation worked great!!

alfopisano (approved):
Reviewed again on m y iOS and everything worked great

tkz00 (changes requested):
* Could we delete methods `IsInProximityRange`, `IsInsideCone` and `IsInArrowLine` from `AimDirection.cs`? 
    * They don’t seem to be used anywhere, except from some commented code in `CustomCharacter.cs`, I deleted the methods and the game ran OK. Are these method going to be used in a future iteration?
* In method `RotatePlayer` of `CustomCharacter.cs`, if the Direction vector has a magnitude of 1, is it necessary to normalize it? I tried deleting it and everything works fine.
* `GetPlayerOrientation`in `CustomInputManager.cs` can also be commented/deleted.
* The changes in `GameServerConnectionManager.cs` seem to be only formatting, do you think it would be good to set formatting rules for the client, and enforce them in the CI like the backend does? We had that in the AFK game, we also had set a personal configuration to auto format c# scripts like we had decided on save. But I think believe it would be better to revert changes to that file in this PR.

client/Assets/Scripts/CustomCharacter.cs
```diff
@@ -42,11 +45,14 @@ protected override void Initialization()
 
     public void RotatePlayer(Direction direction)
     {
-        CharacterOrientation3D characterOrientation = this.GetComponent<CharacterOrientation3D>();
-        characterOrientation.ForcedRotation = true;
         Vector3 movementDirection = new Vector3(direction.X, 0f, direction.Y);
         movementDirection.Normalize();
```
tkz00, line 49: I believe this line could be deleted since `direction` is already normalized.
```suggestion
```

client/Assets/Scripts/Skills/Skill.cs
```diff
@@ -246,49 +247,75 @@ public void EndSkillAnimation(string animationId)
...
+            {
+                vfx.GetComponentInChildren<VisualEffect>().SetFloat("EffectDiameter", diameter);
+            }
+            if (
+                vfx.GetComponentInChildren<VisualEffect>()
+                && vfx.GetComponentInChildren<VisualEffect>().HasFloat("EffectRadius")
+            )
+            {
+                var rotation = this.GetComponent<CustomCharacter>().CharacterRotation();
+                var negative = rotation.x > 0 ? 1 : -1;
```
tkz00, line 310:
```suggestion
```
Could this line be deleted? I don't see use for it.

saugon:
> * Could we delete methods `IsInProximityRange`, `IsInsideCone` and `IsInArrowLine` from `AimDirection.cs`?
>       
>       * They don’t seem to be used anywhere, except from some commented code in `CustomCharacter.cs`, I deleted the methods and the game ran OK. Are these method going to be used in a future iteration?
> 
>     * In method `RotatePlayer` of `CustomCharacter.cs`, if the Direction vector has a magnitude of 1, is it necessary to normalize it? I tried deleting it and everything works fine.
> 
>     * `GetPlayerOrientation`in `CustomInputManager.cs` can also be commented/deleted.

Deleted! 👍🏻 

>     * The changes in `GameServerConnectionManager.cs` seem to be only formatting, do you think it would be good to set formatting rules for the client, and enforce them in the CI like the backend does? We had that in the AFK game, we also had set a personal configuration to auto format c# scripts like we had decided on save. 

Yes, absolutely we need to enforce that (Manu was working on that).

> But I think believe it would be better to revert changes to that file in this PR.

My bad. Done 👍🏻

tkz00 (approved):
Great, works great on my android device, and the code looks good.

alfopisano (approved):
Nice!! Rotation feels great on my iOS

PR #1870, opened 2024-06-27
Add spike feature

### LINKED ISSUE #709: [FEATURE] Bots 🤖
## 🚀 Features
- General design notes: https://www.notion.so/lambdaclass/Bots-9468222e8eba48c187d88aefeb22827e
- [x] Bots need to use the playable characters (`Active` boolean in `characters.json`)
- [ ] ~~They need to either have tailored AI for the different characters and/or understanding of ButtonTypes, or anything else you can come up with so they work with all heroes. New columns can be added to the character/skills JSON to aid this feature as well, if needed~~ We should probably indicate internally in each skill whether it's short/med/long range so the bots have better understanding of when to use them
- [ ] Players that leave should continue to play as bots
- [ ] Prevents bots from stacking on top of each other (they should respect some distance to any target)
- [ ] Avoid moving just in straight lines and somewhat imitate what an input from a real player and joystick would look like
- [ ] They should reduce their engaging chance when under 25% health and try to seek cover (i.e. find a place without many units)
 
## 🐛 Bugs

- [x] Bots need to be made aware of the shrinking circle (battle royale)
- [x] Check if there are other issues that could break their behavior (sometimes they get stuck against walls) that could be a result of the implementation of various features since they were first added

agustinesco:
> [!Important]  
> Merge alongside [This backend pr](https://github.com/lambdaclass/mirra_backend/pull/715)

## Motivation

closes #709

[This backend pr](https://github.com/lambdaclass/mirra_backend/pull/715) added changes to the message.protos file and we need to update it

## Summary of changes

add missing fields to messages.proto


alfopisano (dismissed):
Nice!! Tested on my iOS device and worked great, awesome addition 🙌

alfopisano (dismissed):

client/Assets/Scenes/TitleScreen.unity
AngieDutra, line 1: was it necessary to modify title screen?

client/Assets/Scripts/LineRendererCollider.cs
```diff
@@ -12,7 +12,7 @@ public class LineRendererCollider : MonoBehaviour
 
     void Start(){
         if(Line.positionCount > 0){
-            Smooth();
+            // Smooth();
```
AngieDutra, line 15: If this is commented out is ir relevant to leave it there or should this be deleted?

client/Assets/Scenes/TitleScreen.unity
tkz00: No, that was a mistake.

client/Assets/Scripts/LineRendererCollider.cs
tkz00: That class whole purpose is for testing, I believe leaving that method call commented helps visibility.

We already had this discussion here: https://github.com/lambdaclass/champions_of_mirra/pull/1901#pullrequestreview-2161173606

AngieDutra (approved):
Works awesome!

alfopisano (approved):
Nice!! Tested on my iOS device

alfopisano (approved):
Nice!! Tested on my iOS device

tvillegas98 (approved):

PR #1873, opened 2024-06-27
Fix game timers

alfopisano:
## Motivation
With the bounties screen added we needed to modify the start game timer that affects the countdown used for the players screen and de "survive" animation and the zone shrinking time. For the countdown we did a temporary fix that this PR reverts since now we are doing the corresponding fix on the backend. 

## Summary of changes
Reverted the fix made for the game countdown. 

## How has this been tested?
Test with this [Backend PR](https://github.com/lambdaclass/mirra_backend/pull/739). Also it is better to test this in a normal match rather than a quick game because in those the zone starts disabled. 

Check that the following are working as expected: 
-  Bounties screen and countdown
- Players screen and countdown
- Survive animation
- Movement when battle starts
- Zone timer

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


AngieDutra (dismissed):
Good fix. This is a temporary fix. @Nico-Sanchez will create an issue for this in back repo
Tested in desktop and Android device and works good in both

tkz00 (changes requested):
Te counter seems to be 5 seconds higher than it should be, in all the phases except for the first one.

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/f8af1d7d-e0c7-4556-8545-72b07bea7853

AngieDutra (dismissed):
Tested in Android and desktop. It's working as expected IMO

tkz00 (approved):
Great! Works amazingly on my android!

client/Assets/Scripts/GameServerConnectionManager.cs
```diff
@@ -219,7 +219,8 @@ private void OnWebSocketMessage(byte[] data)
                         gameState.Zone.NextZoneChangeTimestamp - gameState.ServerTimestamp;
                     this.zoneEnabled = gameState.Zone.Enabled;
                     this.gameStatus = gameState.Status;
-                    this.gameCountdown = gameState.StartGameTimestamp - gameState.ServerTimestamp + this.bountyPickTime_ms;
+                    this.gameCountdown = gameState.StartGameTimestamp - gameState.ServerTimestamp;
+                    var position = gameState.Players[this.playerId].Position;
```
AngieDutra, line 223: Question, why not `Position position = gameState.Players[this.playerId].Position;`? instead of using `var`?

alfopisano: This was something that already was on the script when I first opened the PR. I am now checking main and that variable has been deleted so I will delete it also.

manucamejo (approved):

tvillegas98 (approved):
LGTM!

PR #1879, opened 2024-07-01
Fix flow for username with less than 4 characters

### LINKED ISSUE #1815: [BUG] User flow broken if name less than 4 characters
### Steps to Reproduce

1. Open the app like first time opening (i.e. reset PlayerPrefs)
2. Go through guest flow
3. Input a user name with less than 4 letters (e.g. bob)
4. UI is broken

tkz00:
Closes #1815 

## Motivation

There is an error where if you input a username with less than 4 characters, the popup will become transparent, but still be there, blocking the flow of the application, instead of showing the respective error. [[BUG] User flow broken if name less than 4 characters](https://github.com/lambdaclass/champions_of_mirra/issues/1815)

## Summary of changes

- Fixed the issue that made the set name popup transparent instead of displaying an error when a name with less than 4 characters was input.
- Added new logic to the close button of the popup, and a new error message that helps the user better understand the flow.

## How has this been tested?

Clear your player prefs (cache on iOS or android), and set a name, try to edit it, try to put less than 4 characters, save while leaving the input empty, test that flow feels right. Remember it shouldn't let you set usernames with less than 4 characters.

References:
Screenshot_20240701_115021_client
Screenshot_20240701_115029_client
Screenshot_20240701_115100_client

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


alfopisano (approved):
Nice!! Tested on my iOS device and the flow worked great

tkz00:
> I see no point in showing an x button if I'll never be able to use it in that instance. Could we not show it in this case and ave it enabled if the user already has a name set? Screenshot 2024-07-02 at 13 23 30

Agree! The change has been made.

alfopisano (approved):
Nice!! Tested on mi iOS device

AngieDutra (commented):
Now I see the UX better at the beginning but the X option has been lost in the settings once I already have a name. It's still working but I'm not sure this is what we should do either. The X to close the tab should be available in this case, right?
Screenshot 2024-07-02 at 16 25 54

alfopisano (approved):

AngieDutra (approved):
Tested in desktop and Android device. UX issues were fixed. Awesome

AngieDutra:
IMO this is more of a path tho and the real issue remains. Why would everything crash if I have 3 letters in my name? That is the ting to fix also

PR #1880, opened 2024-07-01
[GH-1861] New health bar implementation

### LINKED ISSUE #1861: [FEATURE] New health bar and power up
## 🚀 Feature

This new design integrates the information of the power-ups with the health bar.
In the assets folder, the elements corresponding to the enemy are separated from those corresponding to your information as a player to avoid confusion.

### Reference

reference

Reference for power-ups: (in this case, the numbers only need to have a black shadow)
reference1

### Assets

https://drive.google.com/drive/folders/1r1KxmgHrhk1ZzX1sx1f0_VOG-RKXq9qn?usp=drive_link

### Additional context

Each player should start the match with the lightning icon in the power-up container to avoid displaying a 0. When the player starts obtaining power-ups, the icon should be replaced with the corresponding number.

As a bonus, this design adds:

1. An extra bar indicating the damage the character receives
2. A number accompanying that information

alfopisano:
Closes #1861

## Motivation
This PR aims to implement the new health bar design. It now integrates the power-ups with the health bar and shows the amount of health the character has left. It also modifies the crates health bar. Also now we will be showing how much damage you do the the other player/crate. 

## Summary of changes
Reference: 
reference
crates refe1

NOTE: after talking with Jaz we decided to move the damage text to above the healthbar!!

Reference for the powerUps: 
343859172-1b2b56a9-498c-479b-b057-78a14e715c39

Result: 
<video src='https://github.com/lambdaclass/champions_of_mirra/assets/116388676/468a77b0-1282-49e2-a7e2-0c6d158298a9' />

## How has this been tested?
Just play and take a look at how the health bars are looking and make sure the amount updates correctly. Also test how the damage text is working. 

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


jazimp:
Visual approval ✅

client/Assets/Scripts/HealthBarItem.cs
```diff
@@ -0,0 +1,45 @@
...
+    
+    [SerializeField]
+    Image front,
+        damage,
+        background;
+
+    [SerializeField]
+    MMProgressBar healthBarMM;
+
+    public void setYourHealthBar(ulong health) {
```
AngieDutra, line 25: We start methods with pascal case. This should be `SetYourHealthBar`. I've seen it in several methods added here. I would change it

AngieDutra (commented):
IMO it looks weird to have the health bar so far up in h4ck and almost overlapping muflus. WDYT?
Screenshot 2024-07-04 at 16 20 55
Screenshot 2024-07-04 at 16 21 55

AngieDutra:
I also think we could move up a bit the health bar for valtimer so the hand in his iddle position does't go back and forth and goes through the asset. Otherwise it looks weird
Screenshot 2024-07-04 at 16 25 49

tkz00 (commented):
Two questions
Screenshot_20240705_152527_client

This Uma has a max health of 400, but when on 149, the health bar is more than half full, is this intentional?


https://github.com/lambdaclass/champions_of_mirra/assets/67983386/fee9c9b2-abe3-4ff2-b503-8cc6e9d144ef

And I see the damage values don't match with my health decrease, is this correct?

alfopisano:
> Two questions Screenshot_20240705_152527_client
> 
> This Uma has a max health of 400, but when on 149, the health bar is more than half full, is this intentional?
> 
>  Screen_Recording_20240705_152751_client.mp4 
> And I see the damage values don't match with my health decrease, is this correct?

Related to the bar completion percentage in relation of the actual health the player has I actually do not know if this was intentional. I did not change that logic, only modified the assets. 

In relation with the second question, I actually do not know why the values do not match. What I did for the damage text was showing the difference between the client health (this is the old value) and the server health (the new and actual value). This last value is the one we are updating the healthbar with. Do you have any idea why this could be happening? Maybe I can ask for some help from the backend team.

AngieDutra (dismissed):
I was alive with 0 in my health bar. We should review this and fix it since this implementation may have reveal something wrong in our logic

https://github.com/lambdaclass/champions_of_mirra/assets/82987608/e286f580-bbaf-4e58-90e5-0be315c4ced6

saugon (dismissed):

saugon (approved):
Tested on iphone

tvillegas98 (approved):
Great addition! nice work

PR #1882, opened 2024-07-01
Implement crate dropping delay and power up activation delay

agustinesco:
>[!Warning]
> This pr should be merged alongside [This backend branch](https://github.com/lambdaclass/mirra_backend/pull/735)

## Motivation

We need some time before spawning the power ups when the crates break to play an animation


client/Assets/Scripts/PowerUpsManager.cs
```diff
@@ -33,7 +33,7 @@ public void UpdatePowerUps()
             PowerUp powerUp = powerupEntity.PowerUp;
 
             if (
-                powerUp.Status == PowerUpstatus.Available
+                powerUp.Status == PowerUpstatus.Unavailable
```
tvillegas98, line 36: I understand that if we keep the former check, the power up will not spawn until the `activation_time` has passed right?

tvillegas98 (approved):

client/Assets/Scripts/PowerUpsManager.cs
agustinesco: exactly, this change was made because i encountered that error

tkz00 (approved):
Great work Agus, I feel a longer delay to pick up the power ups could feel better, but that is just configuration work, these changes look and work great! Tested on my android.

manucamejo (approved):

tvillegas98 (approved):
Great job!

PR #1836, opened 2024-06-11
Add new implementations: movement + SAT + reconciliation

manucamejo:
## Motivation

Movement lacks responsiveness, it takes time to follow your inputs.
Client prediction is a mess and innacurate, we have to re-write it.
We need to do the same calculations made by the server's backend (SAT).
After we re-write CP we'll need to implement reconciliation again.

## Summary of changes

- [remove tick references](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/4fd45ef19da77684026cb5dd7d5b5052841efb23)
- [fix interpolation slider](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/334a5f22a81e0206440b7dc7b208f0583951b57e)
- [battle refactor](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/f9ea2aaa4ace41dd7ab167f2a3c3270abe3a10da)
- [UpdatePlayer refactor](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/13f3f5856a5e330e9402e5c584bfbff4d4ea4649)
- [Lower angle threshold. Fix interpolation maximum](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/8f40762bf389ecb0945b8b9107f55f475760c377)
- [Add client prediction option to movement](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/0ab50d73af5859963c2fbd7baca7959290168080)
- [new player movement module](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/8580c4d7a57170ee3c229d9f7ce5043c4926ac2a)
- [Use backend map radius](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/bb0ac1568da3bb00a3b884bf2ccf16338f2ced46)
- [add skills behaviour](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/aa014e8f33ea27372526db424730f9704034a0c8)
- [allow executing skills](https://github.com/lambdaclass/champions_of_mirra/pull/1836/commits/b4f04ee436f40a863ad74735f761a0e1b4a8cd8d)
- Add SAT implementation.
- Add Reconciliation.

## How has this been tested?

We played several matches at the office with the whole team, we concluded this is a way better implementation.
We'll continue developing this in further PRs.

## Checklist
- [x] I have tested the changes locally.
- [x] I have tested the whole game after applying the changes, not only the affected areas.
[...]

PR #1895, opened 2024-07-04
Fix valtimer teleport visuals for other characters

agustinesco:
## Motivation

When we merged the movement rework #1836 we introduced a bug where you can see how other valtimers walk instead of teleporting to the targeted location, this pr fixes that

- [x] Changes tested in an android build


AngieDutra (approved):
Tested on desktop and Android and is working as expected

tvillegas98 (approved):
Tested with Android and Desktop. Great work!

<video src="https://github.com/lambdaclass/champions_of_mirra/assets/67444519/2944b3ca-6318-4844-9c65-673f709e465a">

alfopisano (changes requested):
When using the teleport I noticed that before actually executing it I can see valtimer on the destination point. 
Try to analyze this video frame by frame (tried to upload the video analyzing it but it was too big): 
<video src='https://github.com/lambdaclass/champions_of_mirra/assets/116388676/75b91f1e-c5bd-4f9b-91c2-aa25f71e1e3b' />

PR #1901, opened 2024-07-04
Add spikes to game

tkz00:
> [!CAUTION]
> This PR should be merged alongside backend PR [Spikes positions](https://github.com/lambdaclass/mirra_backend/pull/750)

Closes [Implement Walls](https://github.com/lambdaclass/mirra_backend/issues/709)

## Motivation

Some dynamic obstacles added to the map will be good for the fun of the game. These spike walls can help some players escape or harm them by trapping them.

## Summary of changes

- New assets for the spikes have been added, models, textures, materials, animations, animator controllers and a prefab for them.
- A new module for handling the behavior of the spikes has been integrated on the battle scene, it manages the spikes and their indicator animations based on the state received by the backend.

## How has this been tested?

Enter a match and go to the north east portion of the map, or the south west one, there you should see the new spike obstacles, an indicator will fade in when they are about to rise, if you are on top of them when this happens, then you will be pushed and damaged, in this state, you character will not be able to traverse through them, after some time the spikes will go down underground, allowing players to cross through that path once again. This cycle repeats itself all through the match.

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/d7effcc1-ac77-4f4c-b0e0-b64dec871462

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/67e743e9-247d-46bf-8e69-604546a58ecf

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


client/Assets/Scripts/LineRendererCollider.cs
```diff
@@ -12,7 +12,7 @@ public class LineRendererCollider : MonoBehaviour
 
     void Start(){
         if(Line.positionCount > 0){
-            Smooth();
+            // Smooth();
```
tvillegas98, line 15: Is it okay to remove this line? what it was doing?

client/Assets/Scripts/Obstacles/ObstaclesHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    private void UpdateObstacle(Entity obstacle)
+    {
+        if(this.obstacles[obstacle.Name] != obstacle.Obstacle.Status)
+        {
+            if(mapSpikes.Any(spikes => spikes.name == obstacle.Name))
+            {
+                GameObject inGameSpikes = mapSpikes.Find(spikes => spikes.name == obstacle.Name);
```
tvillegas98, line 41: Maybe we can use `FindIndex` to avoid checking twice the list but it just a comment, not mandatory

client/Assets/Scripts/LineRendererCollider.cs
tkz00: Yes, that whole class servers only to test, it makes lines appear based on the colliders received from the backend, that `Smooth` method smoothes those lines, but for me made it more confusing, that's why I commented it, it doesn't have an effect on the actual game.

client/Assets/Scripts/Obstacles/ObstaclesHandler.cs
tvillegas98: Great!

manucamejo (approved):

tvillegas98 (approved):
Nice work and description @tkz00 !

PR #1910, opened 2024-07-06
Fix effects handling

tvillegas98:
## Motivation

Magic boots trail effect wasn't being deleted

## Summary of changes

This PR fixes that. Now you should be able to apply multiple effects or just use one, in this case the magic boots, and have them deleted when they should

## How has this been tested?

Start a match and test multiple uses of effects and just using the boots

## Checklist
  - [x] Tested in iOS.
  - [x] Tested in Android.


saugon (approved):
Tested on iphone 👍🏻

client/Assets/Scripts/Effects/CharacterMaterialManager.cs
```diff
@@ -52,57 +52,40 @@ public void resetMaterial()
         }
     }
 
-    public IEnumerator applyEffectByKey(MaterialSettingsKey key)
+    public void applyEffectByKey(MaterialSettingsKey key)
```
tkz00, line 55:
```suggestion
    public void ApplyEffectByKey(MaterialSettingsKey key)
```

client/Assets/Scripts/Effects/CharacterMaterialManager.cs
```diff
@@ -52,57 +52,40 @@ public void resetMaterial()
...
-
-        float curent_duration = 0.0f;
-        while (curent_duration <= block.apply_duration)
         {
-            foreach (Renderer renderer in renderers)
-                renderer.material.SetFloat(block.controll_property, curent_duration / block.apply_duration);
-
-            curent_duration += Time.deltaTime;
-            yield return null;
+            block.applyToMaterial(renderer.sharedMaterial);
```
tkz00, line 60: Maybe also change this to ApplyToMaterial, but since that modifies a new file it could be confusing, I leave it to you

client/Assets/Scripts/Effects/EffectCharacterMaterialController.cs
```diff
@@ -46,11 +42,9 @@ private IEnumerator switchMaterialForTime()
         deinit();
     }
 
-    private IEnumerator applyPropertiesForTime()
+    private void ApplyEffectProperties()
     {
-        yield return character_material_manager?.applyEffectByKey(material_settings_key);
-        yield return new WaitForSeconds(duration);
-        yield return character_material_manager?.deapplyEffectByKey(material_settings_key);
+        character_material_manager?.applyEffectByKey(material_settings_key);
```
tkz00, line 47:
```suggestion
        character_material_manager?.ApplyEffectByKey(material_settings_key);
```

client/Assets/Scripts/CharacterFeedbackManager.cs
```diff
@@ -32,28 +32,25 @@ void Awake()
 
     public void ManageStateFeedbacks(Entity playerUpdate, CustomCharacter character)
     {
-        if (playerUpdate.Player.Effects.Count > 0)
+        CharacterFeedbacks characterFeedbacks = character.GetComponent<CharacterFeedbacks>(); // maybe cache this? We can optimize this later.
```
tkz00, line 35: What do you think about adding a conditional to check if `playerUpdate.Player.Effects.Count > 0` before retrieving CharacterFeedbacks?

tkz00 (changes requested):

client/Assets/Scripts/CharacterFeedbackManager.cs
AngieDutra: Then it won't reset the last vfx once it reaches to 0 the effects

client/Assets/Scripts/Effects/CharacterMaterialManager.cs
AngieDutra: done

client/Assets/Scripts/CharacterFeedbackManager.cs
tkz00: Oh yeah! You're right!

tkz00 (approved):
Works great on my android!

saugon (approved):

tkz00 (approved):
Works great on my android!

PR #1919, opened 2024-07-08
Fix h4ck ulti spawning valtimers vfx

tkz00:
When 2 or more players entered a match with characters with pools, the one that entered last would spawn the pool vfx of the earliest when throwing it's own pool skill. This was caused because only one pooler was made for each aoe skill, instead of making a pooler for each skill for each character (how projectile poolers are handled). The issue with this was that when trying to look for the pooler for your character's skill, it would do it by the `skill_id` which is shared betweeen skills of different characters (each basic has `skill_id` 0, each dash is 1 and every ultimate is 2), this meant that if a player tried to cast a pool skill, and there was another player that had entered before him with a pool skill in the same slot (basic, dash or ulti), it would cast that skill.

To fix this I just created a pooler for each skill for each character, and when trying to search for the respective pooler it does so using the `skill_id` as well as the `owner_id`.

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


ncontinanza (approved):
LGTM! Tested in iphone, problem was fixed and everything else worked fine.

alfopisano (changes requested):
Whenever I try to use the ultimate I get all this errors and do not see the vfx: 

<video src="https://github.com/lambdaclass/champions_of_mirra/assets/116388676/a7f0ed84-31ec-4639-9421-e7a7f93f6131" >

alfopisano (approved):
Nice!! Tested playing from my computer and my phone at the same time using valtimer in one device and h4ck on the other and the issue was solved

PR #1920, opened 2024-07-09
Fixed Uma's invisibility bugs

saugon:
## Motivation
Fixed two bugs related to Uma's invisibility:
1. Invisible Umas could see each other.
2. After an enemy Uma came back from invisibility, their health bar didn’t reappear.

## Summary of changes
1. Updated ChracterFeedbackManager: It was using the shared material instead of the instanced one
2. Updated ChracterFeedbackManager and CharacterMaterialManager: The character material was being reset every frame, thereby breaking the logic.

## How has this been tested?
Playing with many Umas and other character and testing the invisibility skill and other effects behavior. 

## Checklist
  - [x] Tested in iOS.
  - [ ] Tested in Android.


tvillegas98 (approved):
Nice catch @saugon !

tkz00 (commented):
One question, shouldn't be `renderer_material_pairs` be cleaned or some of the pairs removed somewhere? Or am I missing some of the logic?

And also when turning invisible with the invincibility effect on it seems to not go away.

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/e7f7ff1d-e289-439f-8d5a-2903dd41909a

tkz00:
> One question, shouldn't be `renderer_material_pairs` be cleaned or some of the pairs removed somewhere? Or am I missing some of the logic?
> 
> And also when turning invisible with the invincibility effect on it seems to not go away.
> Screen_Recording_20240709_000212_client.mp4

I wasn't able to reproduce the second issue like in the video, so I'm not sure what happened.

tkz00 (approved):

PR #1927, opened 2024-07-11
Add two new spikes

tkz00:
> [!CAUTION]
> Needs to be merged alongside [Add two additional spikes](https://github.com/lambdaclass/mirra_backend/pull/756) backend PR.

## Motivation

Current spikes are too far from the center of the map, where most of the action happens.

## Summary of changes

For the spikes to have more impact on the game we've added two new spike walls, they work just like the last ones do.

## How has this been tested?

Play some matches, try to get hit by the new and old spikes, check if the hitboxes feel right, if you can play with another player and check how their character reacts to it.

References:

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/6d312ede-5e40-4d72-a4f2-e0953291287b

https://github.com/lambdaclass/champions_of_mirra/assets/67983386/46d912ad-5f77-430e-8da9-e45b49538a57

DISCLAIMER: There is a known bug where the client prediction behaves differently from the backend, causing the movement and reconciliation to look weird, this is being addressed in a different PR.

https://github.com/user-attachments/assets/e2b0f4ce-ec2c-4265-8dbc-68c2c516c566

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


AngieDutra (changes requested):
What are the changes in battle.unity? Are they necessary? Also, I see this thing which may be easy to solve and better to keep our good practices. The following screen is an override in araban map. It's the only one
Screenshot 2024-07-11 at 12 48 35

Spikes are working fine. Tested in desktop and Android

tkz00:
> What are the changes in battle.unity? Are they necessary? Also, I see this thing which may be easy to solve and better to keep our good practices. The following screen is an override in araban map. It's the only one Screenshot 2024-07-11 at 12 48 35
> 
> Spikes are working fine. Tested in desktop and Android

Yes, my bad, already fixed, just for clarity there are other changes in the `Battle` scene that should be there, but the one from the screenshot, wasn't one of them, actually that was a mistake, the original version is the correct one.

AngieDutra (approved):
Tested on desktop and Android and looks awesome!

alfopisano (approved):
Nice! Tested on my iOS device and worked great

PR #1953, opened 2024-07-19
Add sat fix for invalid axis

agustinesco:
> [!Warning]
> Merge alongside [This backend pr](https://github.com/lambdaclass/mirra_backend/pull/770)

## Motivation

We need that the collision detection an resolution algorithm is the same as the one in the backend, we were missin a fix that avoid players going inside another obstacle and outside of the map when resolving the collision

## Summary of changes

- Added Jeffrey thompson collision detection algorithm
- Added invalid_axis check to SAT algorithm


client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,318 @@
...
+        GameEntity obstacleGameEntity = EntityToGameEntity(obstacle);
+
+        if (gameEntity.shape == "point" && obstacleGameEntity.shape == "circle")
+        {
+            if (pointCircleCollision(gameEntity, obstacleGameEntity)){
+                result.Add(obstacleGameEntity.id);
+            }
+        } else if (gameEntity.shape == "circle" && obstacleGameEntity.shape == "point") {
+            if (pointCircleCollision(obstacleGameEntity, gameEntity)){
+                result.Add(obstacleGameEntity.id);
```
Nico-Sanchez, line 34: All these if's are unnecessary.
I believe it's better to use a switch here, then inside every case, just call the proper function instead of another if, that if could be inside each function and add the entity to the list or not.

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,318 @@
...
+ */
+public static bool LineCircleCollision(GameEntity line, GameEntity circle) {
+    // // Check if the vertices are inside the circle
+    GameEntity point_1 = NewPoint(0, line.vertices[0]);
+    bool inside_1 = pointCircleCollision(point_1, circle);
+    GameEntity point_2 = NewPoint(0, line.vertices[1]);
+    bool inside_2 = pointCircleCollision(point_2, circle);
+    if( inside_1 || inside_2 ){
+        return true;
+    };
```
Nico-Sanchez, line 117: What's inside_1 and inside_2? Please avoid this kind of naming, name the variables after what they represent.

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,318 @@
...
+
+    // Find the closest point on the line to the circle
+    float dist_x = point_1.position.X - point_2.position.X;
+    float dist_y = point_1.position.Y - point_2.position.Y;
+    float line_length = (float)Math.Sqrt((dist_x * dist_x) + (dist_y * dist_y));
+
+    double dot = (((circle.position.X - point_1.position.X)
+        * (point_2.position.X - point_1.position.X))
+        + ((circle.position.Y - point_1.position.Y) * (point_2.position.Y - point_1.position.Y)))
+        / Math.Pow(line_length, 2);
```
Nico-Sanchez, line 127: Is this a dot product? don't call it dot please, that's a different thing.
Don't call it dotProduct either. As I said before, name the variables after what they represent.

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,318 @@
...
+    float uB = ((line_second_vertex.X - line_first_vertex.X)
+        * (line_first_vertex.Y - other_line_first_vertex.Y)
+        - (line_second_vertex.Y - line_first_vertex.Y)
+            * (line_first_vertex.X - other_line_first_vertex.X))
+        / ((otherLineSecondVertex.Y - other_line_first_vertex.Y)
+            * (line_second_vertex.X - line_first_vertex.X)
+            - (otherLineSecondVertex.X - other_line_first_vertex.X)
+                * (line_second_vertex.Y - line_first_vertex.Y));
+
+    return uA >= 0.0f || uA <= 1.0f &&  uB >= 0.0f || uB <= 1.0f;
```
Nico-Sanchez, line 273: This is unreadable. Split the operations and put them in variables named after what they represent, in order to follow the full math operation.
Also uA and uB doesn't tell anything.

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    return false;
+}
+
+public static bool LineLineCollision(GameEntity line, GameEntity other_line) {
+    Position line_first_vertex = line.vertices[0];
+    Position line_second_vertex = line.vertices[1];
+    Position other_line_first_vertex = other_line.vertices[0];
+    Position otherLineSecondVertex = other_line.vertices[1];
```
Nico-Sanchez, line 253: We don't use snake case in C#

Nico-Sanchez (changes requested):
I believe this PR was rushed. I didn't fully reviewed it because I saw a lot of things to fix at first sight.

Please review your own code before leaving it for review.

agustinesco:
> I believe this PR was rushed. I didn't fully reviewed it because I saw a lot of things to fix at first sight.
> 
> Please review your own code before leaving it for review.

Sorry! i meant to leave this pr as a draft, i'm still developing it and i didn't want it to just leave it in my computer so i created a pr

client/Assets/Scripts/CollisionDetection.cs
agustinesco: This is in fact dot product, or at least that's what the author of this algorithm says: 

```
 // get dot product of the line and circle
  float dot = ( ((cx-x1)*(x2-x1)) + ((cy-y1)*(y2-y1)) ) / pow(len,2);
```

client/Assets/Scripts/CollisionDetection.cs
agustinesco: done!

client/Assets/Scripts/CollisionDetection.cs
agustinesco: This is what the original post of line/line collision detection have:
```
  // calculate the distance to intersection point
  float uA = ((x4-x3)*(y1-y3) - (y4-y3)*(x1-x3)) / ((y4-y3)*(x2-x1) - (x4-x3)*(y2-y1));
  float uB = ((x2-x1)*(y1-y3) - (y2-y1)*(x1-x3)) / ((y4-y3)*(x2-x1) - (x4-x3)*(y2-y1));
```

i'm not entirely sure what any of this operations intends but i'll do my best to split them

client/Assets/Scripts/CollisionDetection.cs
agustinesco: force of habit :P as i mentioned i intended to leave this pr as a draft so i'll fix this now

client/Assets/Scripts/CollisionDetection.cs
agustinesco: 1. I ended up using a dictionary to store a tuple containing both entities' shapes as keys and for the value we store a lambda that will call the proper function, I think the readability of this way is good, let me know what you think!
2. hmm idk about changing the collision check behavior, every check function uses another one, and ends up in a long chain of passing the result list between functions, I think the way it's is readable enough. We can talk about this

Nico-Sanchez: 2. Ok, it's a bigger refactor than I expected.

Let's ask about the performance of 1 lambda functions.

client/Assets/Scripts/CollisionDetection.cs
Nico-Sanchez: Yeah I thought so, I understand the math in it.
What I meant is to change the variable name:
> don't call it dot please, that's a different thing.
Don't call it dotProduct either. As I said before, name the variables after what they represent.

For example, it could be:
`float dotProductBetweenLineAndCircle = ( ((cx-x1)*(x2-x1)) + ((cy-y1)*(y2-y1)) ) / pow(len,2);`

Of course this is something I made up, replace the name with what it represents.

client/Assets/Scripts/CollisionDetection.cs
Nico-Sanchez: It's ok. It's hard to name this kind of math. Looks much better now 😄

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,322 @@
...
+    return distance <= circle.radius;
+}
+
+/*
+ * Determines if a collision has occured between a line and a circle
+ * If the distance between the center of the circle and the closest point
+ * of the line is less than the radius of the circle, a collision has occured
+ * Also that closest point should be on the segment
+ */
+public static bool LineCircleCollision(GameEntity line, GameEntity circle) {
```
Nico-Sanchez, line 110: All these collision check functions should be called after an action (with a verb). This way, the boolean returned is easy to understand without checking the function code.

client/Assets/Scripts/CollisionDetection.cs
agustinesco: I think `lineAndCircleDotProduct` is a suitable name

client/Assets/Scripts/CollisionDetection.cs
agustinesco: Every time you declare a lambda c# allocate some memory to save the variable states to run it after, so yeah they'll negatively affect the performance, i restore the if chain since it's not that hard to read and it's performant enough

client/Assets/Scripts/CollisionDetection.cs
agustinesco: Done! i rephrase them to `AreShape1AndSapre2Colliding`

Nico-Sanchez (dismissed):

Nico-Sanchez (dismissed):

Nico-Sanchez:
Sometimes the client and the backend doesn't solve the collision the same way.
We can merge this and fix it in another PR if it's not quick, since this is already a step forward in our fix.

https://github.com/user-attachments/assets/5268a577-09f6-4809-a9c6-212313a6d400

agustinesco:
> Sometimes the client and the backend doesn't solve the collision the same way. We can merge this and fix it in another PR if it's not quick, since this is already a step forward in our fix.
> 
>  Screen.Recording.2024-07-25.at.3.26.44.PM.mov

There's no easy solution for this issue, since this variance in push direction is due to the position in the front

Nico-Sanchez (dismissed):

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,322 @@
...
+            if (AreLineAndCircleColliding(obstacleGameEntity, gameEntity)){
+                result.Add(obstacleGameEntity.id);
+            }
+        } else if (gameEntity.shape == "circle" && obstacleGameEntity.shape == "circle") {
+            if (AreCircleAndCircleColliding(obstacleGameEntity, gameEntity)){
+                result.Add(obstacleGameEntity.id);
+            }
+        } else if (gameEntity.shape == "circle" && obstacleGameEntity.shape == "polygon") {
+            if (AreCircleAndPolygonColliding(gameEntity, obstacleGameEntity)){
+                Debug.Log("collide circle polygon" + obstacleGameEntity.id);
```
tvillegas98, line 51:
```suggestion
```

client/Assets/Scripts/CollisionDetection.cs
```diff
@@ -0,0 +1,322 @@
...
+                result.Add(obstacleGameEntity.id);
+            }
+        } else if (gameEntity.shape == "polygon" && obstacleGameEntity.shape == "circle") {
+            if (AreCircleAndPolygonColliding(obstacleGameEntity, gameEntity)){
+                Debug.Log("collide polygon" + obstacleGameEntity.id);
```
tvillegas98, line 56:
```suggestion
```

tvillegas98 (changes requested):
Great work with this code migration! so sad that we can't fix small difference between client and server.
Left some comments to remove some logs

tvillegas98 (approved):

Nico-Sanchez (approved):

Issue #1796, opened 2024-05-24
[FEATURE] Leaderboard screen


jazimp:
## 🚀 Feature

Implement the new design.

### Reference

![Reference](https://github.com/lambdaclass/champions_of_mirra/assets/134973476/643c9afb-f178-4ce1-8b66-8b1e76175d8d)

### Assets

https://drive.google.com/drive/folders/1sWHKK8zKNo8ndnjb9pXkalKz6y6a2Sqx?usp=drive_link

Buttons: https://drive.google.com/drive/folders/1h93h0kjVentkQGeoLqCXmc7HB0Y7pBGr?usp=drive_link

Icons: https://drive.google.com/drive/folders/18CThUAJRrYy62blR-aP36nF5OzmA_jX-?usp=drive_link

Background: https://drive.google.com/file/d/1KvFyBAm4fWlWDC1Lqvd1sUGJhGAiXOz1/view?usp=drive_link

### Additional context
Disclaimer: As a first iteration, let's have the player reach this screen by tapping the trophy container in the main menu. If testing with the team shows a better path, we'll modify it in a new iteration.

PR #1968, opened 2024-07-22
[GH-1948] Send username to backend

### LINKED ISSUE #1948: [FEATURE] Setting/Changing username on client should reflect that change on the backend
### Motivation

Right now, the username (nickname) of the players is just a player pref cached on the client, the backend holds no record of this information, this has become troublesome since when the backend returns users (like it should for the leaderboard) it displays an auto-incremental username instead of the nickname the player chose. This is also bad since there isn't a way to relate which is the user from the backend of each client.

### Implementation

The backend endpoint for saving the nickname exists already, but it's not being called from the client, we should, on all name setting/changing popups, call that endpoint with the input of the user as a parameter, thus changing the username on the backend.

tkz00:
> [!WARNING]  
> Merge alongside [this backend PR](https://github.com/lambdaclass/mirra_backend/pull/783). Keep in mind using another backend branch will not break this feature, but, this will make the username verification rules not match, and this will cause issues.

Closes #1948 

## Motivation

The username chosen by the player only existed as cache on the client, this caused the backend to use auto-incremental values for the usernames on the database, and when the client retrieved information related to users, such as for the leaderboard, the names didn't show the ones chosen by players.

## Summary of changes

An endpoint to modify the player's username exists in the backend, now, every time the player can set it's username, instead of saving it to the player prefs, it makes a PUT request to that endpoint in the backend, and if the result is successful, only then the username is saved to player prefs, if it's not the error message is show. Since this communication is asynchronous a loading spinner has been added to the main and profile screens for when that request is made.

Also the minimum number of characters for the username has been changed from 4 to 2.

## How has this been tested?

There isn't much difference from the client side, except the loading spinner shown when changing a username, but still the whole game should be tested, try to play some matches with other people and check their names look like they inputted them, although this is not taken from the backend, instead is sent when a user requests to enter a match.

To check the request is correctly made from the client, after setting or changing a username, check that your user in the backend has had it's username updated. To do this while running the backend locally run the command:
```
curl -X GET http://localhost:4001/curse/users/leaderboard
```
and check one of the users has the name you've entered on the client.

To check this you can also run
```
GameBackend.Repo.all(GameBackend.Users.User)
```
on IEx.

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


client/Assets/Scripts/PlayerNameHandler.cs
```diff
@@ -116,4 +174,22 @@ private void Hide()
         this.playerNameHandler.GetComponent<CanvasGroup>().alpha = 0;
         this.playerNameHandler.SetActive(false);
     }
+
+    public void SetLoadingScreen(bool isActive)
```
AngieDutra, line 178: We could re-use the LoadingAnimationController script and avoid doing the same thing in different scripts

AngieDutra (changes requested):

client/Assets/Scripts/PlayerNameHandler.cs
tkz00: Yeah, I didn't see that script, I did unify that logic in [[GH-1796] Add leaderboard scene](https://github.com/lambdaclass/champions_of_mirra/pull/1944), but didn't unify `LoadingAnimationController.cs` since I hadn't noticed that script, the new script that unifies that logic in [[GH-1796] Add leaderboard scene](https://github.com/lambdaclass/champions_of_mirra/pull/1944) is `LoadingSpinner.cs`.

What do you think about me unifying `LoadingAnimationController.cs` and `LoadingSpinner.cs` in [[GH-1796] Add leaderboard scene](https://github.com/lambdaclass/champions_of_mirra/pull/1944) so we don't block this PR and avoid causing merge conflicts for [[GH-1796] Add leaderboard scene](https://github.com/lambdaclass/champions_of_mirra/pull/1944)?

AngieDutra: Sure. Keep in mind LoadingAnimationController is already in use in main. Also it's a more generic name in case our loading animation changes it's still descriptive of the intended use for the script

AngieDutra (approved):
This will be solved in PR #1944 
As for the rest looks good. Tested in Unity desktop and Android build

PR #1982, opened 2024-07-23
[VFX] Bushes Hide effect

saugon:
Based on: https://github.com/lambdaclass/champions_of_mirra/pull/1958

## Motivation

Added bushes with the dither shader, a character shader with the dither feature, and a controller to enable/disable these features using colliders. The colliders can be switched to anything else necessary on the server. this request contains unified character shader.
It’s also good to set up the collision matrix for the new "Hide" layer, as shown in the screen below, to use colliders more efficiently.
зображення

## Summary of changes

Added bushes and their shader, unified character shader, and dither controller.

## How has this been tested?

This effect can be teste via test scene by manualy moving Muflus throgh the bushes

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


saugon:
ℹ️ The logic in CharacterHideController is only for testing purposes, not the actual logic that will be implemented

tkz00 (dismissed):
Great! Works and looks good, although where the shader becomes kind of transparent they look kind of like squares, which I'm not 100% on board with but beyond that it looks good and works fine on android!

Also it was pretty interesting to learn that you can have the camera position as an input for a shader.

client/Assets/Prefabs/Characters/Muflus.prefab
AngieDutra, line 1: A broken script was added to muflus's prefab
Screenshot 2024-07-25 at 17 54 37

client/Assets/Prefabs/Camera/BattleCameraUI.prefab
AngieDutra, line 1: I don't know if we want to add that label in the culling mask of the camera yet since we have yet to implement this feature
Screenshot 2024-07-25 at 17 58 02

AngieDutra (changes requested):

client/Assets/Prefabs/Characters/Muflus.prefab
AngieDutra: Solved this restoring muflus. No need to change this prefab in this PR

client/Assets/Prefabs/Camera/BattleCameraUI.prefab
AngieDutra: TL @saugon expressed there is no need to change this. It won't affect our current game

AngieDutra (approved):
Awesome

tkz00 (approved):

PR #1986, opened 2024-07-24
Add new skill indicators to pre-warm list

AngieDutra:
## Motivation

Added new skill indicators to the prewarmers list to prevent their shader compilation during active game sessions.

## Summary of changes

This PR adds our new skill indicators to prewarmers list

## How has this been tested?

To test this fix, try triggering any skill that requires an indicator. It should not cause a freeze or spike in the Profiler.

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


ncontinanza (approved):
LGTM! Tested locally and on Android, feels much better now.

tkz00 (approved):
Great! We really should stop forgetting about this. Tested on my android.

PR #1991, opened 2024-07-25
[Gh-1964] Refactor sound manager

### LINKED ISSUE #1964: [REFACTOR] Sound manager refactor v2
## 🚧 Refactor
Refactor the current Sound Manager from TopDownEngine to utilize Unity’s built-in audio system. The new implementation should maintain all existing functionality.

AngieDutra:
Closes #1964

## Motivation

How we handled the audio was complicated and wrong

## Summary of changes

This PR:
- Solves UI overrides in both prefabs and scenes
- Refactors how we handle toggle audios
- Refactors how we handle the persistent music in the scenes
- Adjusts how we handle SFX in a more UI / DEV friendly way
Note: this last item is a first iteration. The final goal is to stop using MMSoundManager ( issue #2002 )

## How has this been tested?

Start a match and try all sorts of sound configuration in both main screen and battle. They should persist when changing scenes and be applied as shown in the UI

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


tkz00 (changes requested):
- The main screen music (lobby) doesn't loop, I just tried it in main and it seems to loop (keep in mind there are 27 seconds of no sound at the end of the lobby audio file).
- I feel there is too much of a difference between the first values of the volume and too little for the higher ones. Maybe not for this iteration, or maybe never, but I found this video really interesting that helps to understand this issue (https://www.youtube.com/watch?v=MquQQX0Ak0k)
- I think `VolumeController.cs` could be deleted, I tried it and got no errors.

AngieDutra:
Nice vid!
- Now the music is looped
- I forgot to delete that unused script

AngieDutra:
Yes, @alfopisano keep in mind there are 27 seconds of empty sound at the end of the lobby audio file. This will be fixed by our audio team in the future

tkz00 (approved):
Great! It feels much simpler and better to handle sound now! I tested on android and everything worked!

alfopisano (approved):
Nice!! Tested on my iOS device and worked great

PR #1994, opened 2024-07-25
[GH-1993] Change version text in title screen based of player settings version

### LINKED ISSUE #1993: Show real version on top right of title scree
Now this just shows `VERSION: PRE-ALPHA`

Screenshot_20240725_171403_Champions of Mirra

tkz00:
Closes #1993 

## Motivation

First iteration to show the version of the application, to keep track of the changes that are available.

## Summary of changes

Retrieve the version registered on the player settings with `Application.version` and show it on top right of title screen.

## How has this been tested?

Enter to title screen, check the version is the same one you can find in the `Version` field in player settings of the unity editor, then go past the title screen to the main screen to check nothing breaks.

Reference (that's just a testing value):

Screenshot_20240725_173116_client

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


client/Assets/Scripts/UI/TitleScreenController.cs
```diff
@@ -38,6 +43,10 @@ public class TitleScreenController : MonoBehaviour
 
     void Start()
     {
+        StringBuilder sb = new StringBuilder();
```
AngieDutra, line 46: Is StringBuilder the best approach or an overkill in this particular text? I see the potential for other dynamic texts we have but I don't know if there is a performance issue here for us to use this. Or if we are suppose to use it always. What do you think? Was this a conversation with our TL?

AngieDutra (approved):
I tested this and worked as expected

client/Assets/Scripts/UI/TitleScreenController.cs
saugon: In this situation is an overkill. _StringBuilder is designed to optimize performance when performing multiple string manipulations, especially in loops or when concatenating many strings_.

Let's change it to use the + operator 👍🏻

saugon (approved):
Tested on iOS 👌🏻

manucamejo (approved):

PR #2000, opened 2024-07-29
Add spikes VFX

AngieDutra:
## Motivation

Our in-game spikes were missing vfxs

## Summary of changes

This PR:
- adds a test scene called DynamicObstacle_test_scene
- adds those vfxs to the in-game spikes (start / idle / stop)

## How has this been tested?

Start a match and see the vfx applied. Test performance and implementation logic. Also test how it renders in mobile

To keep into account:
- The spike starts first without VFX because they are hardcoded in the client to show. This bug is in main and separate from the accurate VFX implementation. The issue for this is #2001 
- I wanted to replace the coroutine animation with dotween but it turns out it would be too complicated to add that here and it's low priority. This is also in issue #2001 

## Checklist
  - [ ] Tested in iOS.
  - [x] Tested in Android.


tkz00 (approved):
Everything looks good and works on my android

alfopisano (approved):
Works great and looks amazing on my iOS device

Issue #1947, opened 2024-07-17
[FEATURE] Add animation to indicate auto-selection of bounty


AngieDutra:
### Motivation

We don't have in client a way to show the client a bounty was automatically selected if you have failed to choose one in time

### Implementation

The animation and assets needed for this implementation must me decided alongside @marco-paulucci @saugon @jazimp and the team

### Additional context
This will be needed after PR #1941 is merged

PR #2008, opened 2024-08-01
Fix default bounty crashing game

agustinesco:
> [!warning]
> Should be merged alongside [This backend branch](https://github.com/lambdaclass/mirra_backend/pull/828)

## Motivation

When not selecting a bounty at the start of a match the backend at random assigns you one, if that bounty was completed the client breaks.

## Summary of changes

New message from the backend has been added `BountySelected`, when this message is received by the client the bounty it contains is set as the selected bounty, previously it was set only when a bounty was selected on the prepare for battle animation. This selected bounty is used to display the state of the bounty at the end splash and to increase the amount of coins of the user visually on the client. This broke the client since it was trying to access these values on a null bounty.

## How has this been tested?

Play some matches, try both to not select and select bounties, finish matches completing and not completing them, see that every flow works as expected and that nothing breaks.

DISCLAIMER: This is just a fix so the client won't break, because of this the bounty selected at random by the backend isn't communicated to the player in any way, there is an issue open for this [[FEATURE] Add animation to indicate auto-selection of bounty](https://github.com/lambdaclass/champions_of_mirra/issues/1947).

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


Nico-Sanchez (approved):

client/Assets/Scripts/Battle.cs
```diff
@@ -647,7 +647,10 @@ long pastTime
             {
                 bountyCompletionManager.GenerateBountyCompletionPopUp();
                 bountyCompletionManager.UpdatePlayerCurrencies();
-                MainScreenManager.bountyReward = (int)GameServerConnectionManager.Instance.bountySelected.Reward.Amount;
+                if(GameServerConnectionManager.Instance.bountySelected != null)
```
tvillegas98, line 650:
```suggestion
                if(GameServerConnectionManager.Instance.bountySelected)
```
Isn't it the same? just wondering, not mandatory comment

tvillegas98 (approved):

Issue #801, opened 2023-08-21
[FEATURE] D'agna


marco-paulucci:
### Skills

- [ ] #796 
- [ ] #797
- [ ] #798
- [ ] #799
- [x] #800 

### Character
- [x] #556 
- [x] #749

PR #2011, opened 2024-08-05
Filter map generated pools

tkz00:
> [!WARNING]  
> While this doesn't have to be merged alongside [[GH-801] Add slow fields as pools in game configuration](https://github.com/lambdaclass/mirra_backend/pull/803/files), it is needed to test this branch.

Since the backend has added new types of obstacles, including oil pools, which are of type pool, the client is breaking since the `PoolsHandler.cs` was trying to instantiate prefabs for those pools searching that prefab via a `SkillId`, since these new oil pools have "0" as `SkillId` indicating they don't have a skill, we filter out the pools from the backend with that value to avoid crashing. Also try out all the character pool skills (right now only H4ck and Valtimer have one), having multiple H4ck's and Valtimer's on the same match and throwing those skills repeatedly.

## Checklist
  - [ ] Tested in iOS.
  - [ ] Tested in Android.


client/Assets/Scripts/Pools/PoolsHandler.cs
```diff
@@ -42,7 +42,7 @@ public void CreatePoolsPoolers(List<SkillInfo> skillsInfo)
 
     public void UpdatePoolsActions()
     {
-        List<Entity> poolsStates = GameServerConnectionManager.Instance.gamePools;
+        List<Entity> poolsStates = GameServerConnectionManager.Instance.gamePools.Where((poolState) => poolState.Pool.SkillKey != "0").ToList();
```
tvillegas98, line 45: This seems odd but okay!

tvillegas98 (approved):

AngieDutra (commented):
I can't test this since this logs make it hard for me to even do a simple shoot or even dash or move
Screenshot 2024-08-06 at 12 29 21

alfopisano (approved):
Nice!! Tested on my computer and iOS device using localhost and brazil and worked great

PR #2012, opened 2024-08-06
New map layout

tkz00:
> [!WARNING]  
> Needs to be merged along with [this backend PR](https://github.com/lambdaclass/mirra_backend/pull/836)

We wanted to add new elements to the map, to make it more interesting and add variability from match to match. This is just a first iteration that moves spikes to different positions and adds bushes that hide players and two kinds of pools, one that make characters slower and other that allows projectiles to pass through them (currently they are sand and water respectively but that could change).

## How has this been tested?

Play some matches with other players, interact with all the new elements on the map, check that their hit boxes match with what is seen on screen and that all behavior works as expected, bushes should hide you (50% opacity), make enemies invisible when far away and transparent (50% opacity) when close and both on the bush. Water pools should let projectiles pass through but not characters and sand pools will slow characters.

Reference:
Araban_v0 1

In Game:
Screenshot 2024-08-02 at 4 53 38 PM

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


alfopisano (approved):
Nice!! Tested locally and on my iOS device and worked great

AngieDutra (approved):
Nice, tested in Android and playtested it with brasil. IMO it is missing some correlation to backend data. Some things show in off positions than the backend config says so. They are similar but not the same. However, I talked about this with theo and he said it will be adjusted later on in another PR. I have no more comments other than that

PR #2017, opened 2024-08-07
[GH-2015] Compare versions with backend

### LINKED ISSUE #2015: Compare to backend version on start of application
We want to check that backend and client versions match, to avoid playing with different ones, which can cause unexpected behavior. To do this, on the start of the game, a request to the backend will be made for the version number (the format of both will be XX.XX.XX). For this first iteration, we'll only check that the middle number of the backend is not higher than the middle number of the client. If this is the case, then the user should be prompted of this problem and not be allowed to continue beyond the title screen. This check will be repeated whenever the user changes the server.

tkz00:
Closes #2015 

## Motivation

Discrepancies between server and client can cause unexpected behaviors, this is why playing with outdated clients should be impossible. 

## Summary of changes

- Retrieve the server version when trying to log in (when guest or google sign in buttons are pressed).
- Compare only the middle number of the version from the server to the one on the client xx.XX.xx.
  - If the client's version is equal or higher to the server one the application continues it's flow as always.
  - If the client's version is lower than the server's, then a popup is shown informing of this problem and the user continues in the title screen instead of moving to the main screen.

## How has this been tested?

Try changing the version in the client to test the different situations and their respective outcomes, seeing the popup and not being able to access the main screen if the client version is lower than the server's and having a normal flow if the versions are equal or the client's is higher. Also try to test this both on a production server and on local.

You can change the client's version in `Edit/Project Settings` in the `Version` field

Screenshot 2024-08-07 at 3 09 51 PM

## Checklist
  - [ ] Tested in iOS.
  - [X] Tested in Android.


alfopisano (approved):
Great!! Tested locally and on my iOS device with brazil and everything worked as expected

tvillegas98 (approved):

